PR #3, opened 2024-10-10
Add 404 page

juliasilge:
Adds a 404 page, so we don't see the default Netlify 404.

### DIFF
--- a/404.qmd
+++ b/404.qmd
@@ -1,0 +1,6 @@
+---
+title: "Page Not Found"
+format: html
+---
+
+Oops! The page you requested isn't available. Try searching our documentation or find out [how to get started](start.qmd) with Positron.
--- a/404.qmd
+++ b/404.qmd
@@ -3,4 +3,4 @@ title: "Page Not Found"
 format: html
 ---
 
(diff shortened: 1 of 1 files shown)


404.qmd
```diff
@@ -0,0 +1,6 @@
+---
+title: "Page Not Found"
+format: html
+---
+
+Oops! The page you requested isn't available. Try searching our documentation or find out [how to get started](start.qmd) with Positron.
```
juliasilge, line 6: I'd love to use a link here to bring up the search bar, but I don't know if that is currently possible. I'd want to do something like:

```
Try [searching our documentation](some-useful-link-that-brings-up-search) or find out [how to get started](start.qmd) with Positron.
```

AshleyHenry15 (approved):
tested locally - looks good! We will want to test again once it is live - I am not sure how this will work/if there is an additional step that we need to take (the docs 404 page is coming from the s3 bucket so I am not that knowledgable about other quarto sites and the 404 page.

juliasilge:
Thank you for showing me a way to get the search results to pop up via a link in #4 @AshleyHenry15! 🙌

PR #9, opened 2024-10-23
Use a less alarming color for the announcement banner

petetronic:
The type danger uses red and seems alarming. Trying out warning which should be less red.


juliasilge (approved):
Thanks!

PR #12, opened 2024-11-01
Add FAQ on Positron with a server

juliasilge:
Let's keep this as draft until we have Positron on Workbench!


juliasilge:
With [Workbench Version 2024.12](https://docs.posit.co/ide/server-pro/) now released, let's merge in this FAQ to point folks to.

PR #22, opened 2024-11-19
Font color update

AshleyHenry15:
Adding a new font color for the website, using the same color as Shiny docs.


AshleyHenry15:
I am okay with the posit-dark-gray-1 change and the font weight for the body font, but I would rather that the rest be ignored for now. We did a lot of testing and the other items seem unnecessary until we can develop a more solid/approved plan moving forward.

juliasilge:
Thank you so much @AshleyHenry15! In my testing `-webkit-font-smoothing: antialiased` really does make _quite_ a difference (especially for some of my stakeholders who are on Linux and such) so I am going to keep that one for now as well.

I am very happy to help in whatever way is best for a more long term, approved plan (just let me know! happy to work on writing up current limitations or similar) and am glad to have this stopgap solution for now.

PR #23, opened 2024-11-20
Add advice on Rtools

juliasilge:
For context, Rtools is a Windows requirement for mainly R package developers. Maybe I should more strongly advise toward rig, given how easy it is? I don't feel like we can _only_ tell folks about rig here.

Related to posit-dev/positron/issues/5358 but we will eventually want to do a bit more work in the product itself related to Rtools.

### DIFF
--- a/start.qmd
+++ b/start.qmd
@@ -28,6 +28,12 @@ Before installing Positron, ensure your Python and/or R environments are ready t
 
 If you're using Windows, make sure you have the [latest Visual C++ Redistributable](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#latest-microsoft-visual-c-redistributable-version) installed.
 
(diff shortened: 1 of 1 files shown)


start.qmd
```diff
@@ -28,6 +28,12 @@ Before installing Positron, ensure your Python and/or R environments are ready t
 
 If you're using Windows, make sure you have the [latest Visual C++ Redistributable](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#latest-microsoft-visual-c-redistributable-version) installed.
 
+If you're an R user, note that Positron doesn't currently bundle [Rtools](https://cran.r-project.org/bin/windows/Rtools/rtools40.html). If you need Rtools for your package development or other work, you can either use the official guidance from CRAN on installing Rtools and putting it on the path, or alternatively, use [rig](https://github.com/r-lib/rig) to install and set up Rtools:
```
DavisVaughan, line 31:
```suggestion
If you're an R user, note that Positron doesn't currently bundle [Rtools](https://cran.r-project.org/bin/windows/Rtools/). If you need Rtools for your package development or other work, you can either use the official guidance from CRAN on installing Rtools and putting it on the path, or alternatively, use [rig](https://github.com/r-lib/rig) to install and set up Rtools:
```

Slightly incorrect link for Rtools. That points to an old version. Here is a more generic link that lets the user choose their Rtools version based on the R version they have (but again, rig would have handled that for them)

DavisVaughan (approved):
I think this is probably the best we can do rn

PR #26, opened 2024-12-03
New downloads page

juliasilge:
This PR starts our new page for providing downloads via the `https://cdn.posit.co/positron/` URLs.

In the very first initial here, I have this set up via Quarto [parameters](https://quarto.org/docs/computations/parameters.html) but we can make this as complicated as we want moving forward, in terms of older releases being available and such.


juliasilge:
I set up Plausible analytics for downloads in 28533ee2f5e706feb56a600ac6112ddd1f39a62e, and according to them, it is all set up correctly.

juliasilge:
Some next steps after this include:

- make a button based on the user's OS, like what [Quarto has](https://quarto.org/docs/get-started/)
- start a page for release notes (see release notes for [Quarto](https://quarto.org/docs/download/) which are one page per release and [Workbench](https://docs.posit.co/ide/news/) which are one big page with a heading per release)

petetronic (approved):
Tested out the preview site and the download links looked good. As Julia mentioned, in a future change we'll have a convenient button, but this is a great first step.

PR #28, opened 2024-12-10
Add more prominent link to GitHub repo in navbar

juliasilge:
@sharon-wang pointed out that there is not a prominent link to point folks to our GitHub repo from the homepage or top nav.

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -45,6 +45,7 @@ website:
         href: troubleshooting.qmd
     right:
       - icon: github
+        aria-label: 'Link to Positron Github Repository'
         href: https://github.com/posit-dev/positron
 
   sidebar:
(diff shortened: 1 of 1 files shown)


_quarto.yml
```diff
@@ -43,6 +43,9 @@ website:
         href: interpreters.qmd
       - text: "Help"
         href: troubleshooting.qmd
+    right:
+      - icon: github
+        href: https://github.com/posit-dev/positron
```
sharon-wang, line 48: Since we're using an icon for the link (with no visible text label), let's include an aria-label:

```suggestion
    right:
      - icon: github
        aria-label: 'Link to Positron Github Repository'
        href: https://github.com/posit-dev/positron
```

sharon-wang (approved):
LGTM! I navigated to the icon with VoiceOver and "Link to Positron Github Repository" was read out.

Thank you for adding this :D

PR #29, opened 2024-12-12
Add Windows instructions on launching Positron from the terminal

petetronic:
The `Install 'positron' command in PATH` is available on macOS. For Windows, the installer includes an option to do this for you. Adding instructions to note this option as well as how to do this manually if you did not select this option during installation.


juliasilge (approved):
Thank you so much! 🙌

start.qmd
```diff
@@ -62,10 +62,26 @@ pak::pak(c("usethis", "cli", "crayon", "rlang", "roxygen2", "pkgload"))
 install.packages(c("usethis", "cli", "crayon", "rlang", "roxygen2", "pkgload"))
 ```
 
-## Add Positron to your path
+## Launch Positron from a terminal
 
-After you have [installed Positron](download.qmd), you can add Positron to your path via the command *Install 'positron' command in PATH*:
+If Positron is on your path, you can launch it from the terminal. Type `positron .` to open Positron in the current folder, or `positron foo` to open Positron in a custom folder of your choice.
```
AshleyHenry15, line 67:
```suggestion
If Positron is on your path, you can launch it from the terminal: 

- Type `positron .` to open Positron in the current folder, _or_ 
- `positron foo` to open Positron in a custom folder of your choice.
```

AshleyHenry15: Just a suggestion - I like to pull steps/options out into a list to make it more scalable if we are offering multiple options/routes in a sentence/paragraph.

start.qmd
```diff
@@ -62,10 +62,26 @@ pak::pak(c("usethis", "cli", "crayon", "rlang", "roxygen2", "pkgload"))
...
+
+See the instructions below on how to add Positron to your path based on operating system.
+
+### macOS
+
+After you have [installed Positron](download.qmd), you can add Positron to your path via the Command Palette. Press `Cmd+Shift+P` to open the palette and type *Install 'positron' command in PATH*.
```
AshleyHenry15, line 73: Quarto/markdown has an accessibility issue with a screen reader; I don't think that it will recognize this link as a "link". It may only happen with anchor links like [link name](#heading-name) but I think it may have similar behavior. Just wanted to call it out.

start.qmd
```diff
@@ -62,10 +62,26 @@ pak::pak(c("usethis", "cli", "crayon", "rlang", "roxygen2", "pkgload"))
...
 
 ![Add Positron to your path](images/positron-path.png)
 
-After you have executed this command once, Positron will be on your path and then you can use `positron .` in a directory to open Positron in that folder, or `positron foo` to open a folder located at your current path.
\ No newline at end of file
+After you have executed this command once, Positron will be added to your path. Restart your terminal to pick up the changes to your path.
```
AshleyHenry15, line 77:
```suggestion
After you have executed this command once, Positron is added to your path. Restart your terminal to pick up the changes to your path.
```

start.qmd
```diff
@@ -62,10 +62,26 @@ pak::pak(c("usethis", "cli", "crayon", "rlang", "roxygen2", "pkgload"))
...
+
+### Windows
+
+Positron installers on Windows include an option to add a `positron` command to your path. This installation option is selected by default.
+
+![Windows installer option Add to PATH](images/installer-windows-add-to-path.png)
+
+If you had unchecked this option during installation, you can manually add the `bin` folder of your Positron installation to your path. For a System level installation, this folder is located at `C:\Program Files\Positron\bin` by default. For a User level installation, this folder is located at `C:\Users\username\AppData\Local\Programs\Positron\bin` by default. You will need to restart your terminal after updating your path.
```
AshleyHenry15, line 85:
```suggestion
If you had cleared this option during installation, you can manually add the `bin` folder of your Positron installation to your path. For a System level installation, this folder is located at `C:\Program Files\Positron\bin` by default. For a User level installation, this folder is located at `C:\Users\username\AppData\Local\Programs\Positron\bin` by default. You will need to restart your terminal after updating your path.
```

Style guide:
Don't use "uncheck" - use "clear"
https://positpbc.atlassian.net/wiki/spaces/ENG/pages/594509972/Posit+Documentation+Style+Guide

start.qmd
```diff
@@ -62,10 +62,26 @@ pak::pak(c("usethis", "cli", "crayon", "rlang", "roxygen2", "pkgload"))
...
+
+Once Positron is on your path you can use a terminal to launch Positron.
```
AshleyHenry15, line 87:
```suggestion
Once Positron is on your path, you can use a terminal to launch Positron.
```

AshleyHenry15 (approved):
Minor suggestions

start.qmd
jennybc, line 67:
```suggestion
If Positron is on your `PATH`, you can launch it from the terminal. Type `positron .` to open Positron in the current folder, or `positron foo` to open Positron in a custom folder of your choice.
```

Maybe this makes it more clear that we really mean 'on your `PATH`', with the technical connotations that has? I see this phrase appears many times, so obviously we want to be consistent.

AshleyHenry15: Anything that we are reusing multiple times - we should insert using an include

petetronic: @jennybc yeah I had wondered if these should be PATH or `PATH`, though went with the existing generic path usage. I think PATH capitalized matches the nature of it being an env var. @AshleyHenry15 I see Quarto has the concept of an include, though would that be used for a single word used over and over?

start.qmd
petetronic: I see, I just moved this link from existing content. The href shows up as `https://deploy-preview-29--positron-posit-co.netlify.app/download.html` in the preview, could you show me how to test the screen reading issue?

start.qmd
AshleyHenry15: If can be used for single lines of text, chunks/paragraphs, code blocks, etc.

https://github.com/rstudio/rstudio-pro/blob/main/docs/user/workbench/positron/guide/proxying-web-servers.qmd

So - you can create a _snippets directory (if you don't already have one).

create a file - `launch-fr-terminal.qmd` (or whatever)

Add the block of content that you want to reuse:
If Positron is on your `PATH`, you can launch it from the terminal. Type `positron .` to open Positron in the current folder, or `positron foo` to open Positron in a custom folder of your choice.

Wherever you want to include the content, do so with 

`{{< include _snippets/launch-fr-terminal.qmd >}}`

start.qmd
AshleyHenry15: Are you running on a Mac? I use VoiceOver

start.qmd
juliasilge: I may be misunderstanding but I don't think the idea of using `include` applies to the original question here.

I am neutral on styling as path vs. `PATH` and tend to think either is pretty clear; if someone believes that one is more helpful than another, happy to go with that!

AshleyHenry15: I assumed that there was a need based on:
>I see this phrase appears many times, so obviously we want to be consistent.

Whatever you decide, if we can capture it in an include so it is consistent across the docs/file, that is ideal and easier to maintain.

start.qmd
petetronic: Ok, I used VoiceOver, and it reads it out as a link and says the text "installed Positron". While I could improve the phrasing here, or look at aria-labels, I think I'm going to just remove this link as it isn't necessary.

PR #31, opened 2024-12-19
fix typo amd64 --> arm64

sharon-wang:
I think we might mean arm64 here? If this isn't actually a typo, let me know!


jmcphers (approved):
Yes, thank you for catching this!

PR #36, opened 2025-01-21
Add WSL to Remote SSH page

kv9898:
Currently, Positron does not *officially* support connecting to WSL. However, I have built a fork of the Open Remote - WSL extension for Positron and have been using it regularly without known problem. The extension has been downloaded almost 700 times from Open VSX Registry. I therefore think it is worth informing new users of Positron that a community-maintained WSL functionality is available, and they can inspect its source code and use it at their own risk.

### DIFF
--- a/remote-ssh.qmd
+++ b/remote-ssh.qmd
@@ -1,10 +1,10 @@
 ---
-title: "Remote SSH and WSL"
+title: "Remote SSH"
 ---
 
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
Thank you so much for this contribution, @kv9898! We are happy to point folks to the WSL extension, especially since the extensibility options for Positron are a big part of why we have chosen to build on this architecture. However, I don't think we're quite comfortable with taking WSL off that list of "not currently supported by us" so I have moved your helpful info to a footnote with some clarification.

jmcphers (approved):

PR #38, opened 2025-01-28
Document the 'focus variables' command

jmcphers:
Documents the new "Focus on Variables View" keybinding added in recent builds.


juliasilge (approved):
Thank you so much!

PR #40, opened 2025-02-09
Update remote SSH screenshot for new setting name

juliasilge:
Addresses the last bit of https://github.com/posit-dev/positron/issues/5824 with the one screenshot that has an updated setting name in it.

@jmcphers I couldn't get the screenshot to look _exactly_ the same so if you think this isn't quite the same enough, definitely feel free to replace my screenshot here in this PR.


jmcphers (approved):
LG, thank you for updating this!

PR #47, opened 2025-02-18
Add agree to privacy policy to download page

timtmok:
Address https://github.com/posit-dev/positron-builds/issues/224

Adds checkbox agreeing to privacy policy before showing the downloads. The checkbox must be selected before the download links appear.

The checkbox has a click handler that toggles a `hidden` class attribute that hides the element.

https://github.com/user-attachments/assets/43e9c129-c068-467c-a614-a4d44fffcf8a


juliasilge (commented):
Thank you so much for working on this! 🙌 

What do you think our options would be if we invested just a little bit more in the UX experience? To me, this version feels quite jumpy ("SURPRISE, HERE IS THE TABLE!!!") and a bit unusual for typical website UX.

- Instead of totally hidden, could the table be grayed out or obscured/unclickable?
- What about a popup that must be accepted?
- Other ideas?

I'm wondering if we can get an experience that may be less surprising to typical website users with a bit more time investment.

timtmok:
> Thank you so much for working on this! 🙌
> 
> What do you think our options would be if we invested just a little bit more in the UX experience? To me, this version feels quite jumpy ("SURPRISE, HERE IS THE TABLE!!!") and a bit unusual for typical website UX.
> 
> * Instead of totally hidden, could the table be grayed out or obscured/unclickable?
> * What about a popup that must be accepted?
> * Other ideas?
> 
> I'm wondering if we can get an experience that may be less surprising to typical website users with a bit more time investment.

These ideas are helpful! I wasn't sure what the experience should be like. I did have something to blur the table. Hiding it was easiest since the blur still requires actual disabling of the links. I'll add some link disabling and that will open up options for how the content is revealed.

timtmok:
@juliasilge I've added a change with a couple of things to try out once the preview deployment has completed.

* The links are greyed and disabled until the checkbox is checked
* Once the checkbox is checked, the links are normal and clickable again
* Clicking on a link without the checkbox checked will show a dialog

The dialog can be customized more for style and behaviour. It's possible to change the dialog buttons to accept/decline the license and privacy and only allow the download if accept is selected.

I think with some more changes the dialog is more obvious why the download didn't start and can present the option of accepting or declining.

juliasilge:
This is looking good! I think not having the table appear/disappear (but instead be disabled) is much better.

> I think with some more changes the dialog is more obvious why the download didn't start and can present the option of accepting or declining.

Yes, I agree that the dialog is more obvious and a few more changes here would be great:

- Can we style the dialog to inherit the styling of the site overall and change its placement? Much like the cookie consent support that Quarto has (see [an example in the wild](https://www.gregdubrow.io/posts/quarto-blogs-and-gdpr/) here)
- Is it possible to use the dialog as the default/only way that people agree, and show the dialog when we load the page instead of when someone clicks?

download.qmd
```diff
@@ -2,13 +2,28 @@
...
 Please review [Positron's license agreement](licensing.qmd) and [privacy policy](https://posit.co/about/privacy-policy/). Your acceptance of this license agreement and privacy policy is required as a condition to proceeding with your download or use of the software.
 
-```{r}
+<div id="privacy-policy-agreement">
+  <input type="checkbox" id="privacy-checkbox" onclick="privacyClickHandler(this)">
+  <label>I agree to the [Positron license agreement](licensing.qmd) and [Posit Privacy Policy](https://posit.co/about/privacy-policy/).</label>
+</div>
+
+<dialog id="accept-dialog">
+<p>Agree to the privacy policy to download Positron.</p>
```
petetronic, line 22: Minor, but can we sync on this wording?

timtmok: Yes, I've updated the dialog text and used the text that was for the checkbox.

_download_modal.html
```diff
@@ -0,0 +1,20 @@
+<div id="accept-dialog" class="modal" tabindex="-1">
+    <div class="modal-dialog modal-dialog-centered">
+        <div class="modal-content">
+            <div class="modal-header">
+                <h5>License Agreement and Privacy Policy</h5>
+                <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+            </div>
+            <div class="modal-body">
+                <p>
+                    I agree to the <a href="licensing.html" target="_">Positron license agreement</a> and
```
timtmok, line 10: I tried using `href="licensing"` here but the preview can't navigate to it like with a live site. I've set the target to `licensing.html` so that it works in the preview.

timtmok:
It's a minor difference with it being a dialog now instead of a checkbox but I think the desired behaviour is the same. The user is agreeing by clicking on `Accept` on the dialog.

juliasilge:
I am not seeing a dialog at <https://deploy-preview-47--positron-posit-co.netlify.app/download.html> now; do you know what might have happened?

_download_modal.html
```diff
@@ -0,0 +1,20 @@
+<div id="accept-dialog" class="modal" tabindex="-1">
+    <div class="modal-dialog modal-dialog-centered">
+        <div class="modal-content">
+            <div class="modal-header">
+                <h5>License Agreement and Privacy Policy</h5>
+                <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+            </div>
```
juliasilge, line 7: This seems a bit redundant with the modal content. What do we think about removing the header?

timtmok: image

Here's what it looks like without the modal header. I think this is much cleaner.

juliasilge: Yes, I like this a lot! 🎉 

I'm not seeing this dialog at https://deploy-preview-47--positron-posit-co.netlify.app/download.html, including in incognito mode. Do you know what might be happening there?

timtmok: I got the details after a screenshare with Pete. The modal doesn't appear until clicking on a link to download. It was the first thought for me when imagining how it it work. It's not too much to change it to show on page load.

Another option is reverting to the checkbox with a dialog showing when attempting to download without agreeing. So the user clicks the link and if the checkbox isn't checked, a dialog appears preventing the download with a reason.

timtmok:
Here's an alternate option with a fade to show that the content is available. The links are not clickable or focusable and the colour is greyed out until the checkbox is ticked.

https://github.com/user-attachments/assets/9fff5990-2b12-4511-be4d-7d6caf473042

juliasilge (approved):
I am happy with the experience we landed on here (callout block plus checkbox). 👍

timtmok:
This is that last change's look

image

PR #49, opened 2025-02-24
Update Python prerequisites given ipykernel bundling

seeM:
Related to https://github.com/posit-dev/positron/issues/6233.

### DIFF
--- a/start.qmd
+++ b/start.qmd
@@ -38,9 +38,16 @@ rig system rtools add
 
 Positron works with [actively supported versions](https://devguide.python.org/versions/#versions) of Python, from 3.8 to Python 3.12. We recommend [pyenv](https://github.com/pyenv/pyenv) and [pyenv for Windows](https://github.com/pyenv-win/pyenv-win) for managing Python versions. On Linux, you'll want to [be aware of the build requirements](https://github.com/pyenv/pyenv/wiki/common-build-problems); for example, you'll need the system library for SQLite (`sqlite-devel` or `libsqlite3-dev`) installed ahead of time so pyenv can properly build your chosen Python version for use with Positron.
 
-Positron requires the IPykernel package (version 6.19.1 or higher) to communicate with Python. IPykernel is bundled with Positron for the CPython implementation, which is the default and most widely used Python implementation.
+Positron requires the [IPython kernel](https://pypi.org/project/ipykernel/) (version 6.19.1 or higher) to communicate with Python. By default, the IPython kernel is bundled with Positron for the CPython implementation (the standard Python implementation) and added to the [Python search path](https://docs.python.org/3/library/sys_path_init.html).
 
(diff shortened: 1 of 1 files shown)


start.qmd
```diff
@@ -38,14 +38,14 @@ rig system rtools add
 
 Positron works with [actively supported versions](https://devguide.python.org/versions/#versions) of Python, from 3.8 to Python 3.12. We recommend [pyenv](https://github.com/pyenv/pyenv) and [pyenv for Windows](https://github.com/pyenv-win/pyenv-win) for managing Python versions. On Linux, you'll want to [be aware of the build requirements](https://github.com/pyenv/pyenv/wiki/common-build-problems); for example, you'll need the system library for SQLite (`sqlite-devel` or `libsqlite3-dev`) installed ahead of time so pyenv can properly build your chosen Python version for use with Positron.
 
-Positron communicates with Python via the IPykernel package, so it needs to be installed for the Python environment you want to use with Positron. If you are managing your Python project using a tool like venv or conda, add `ipykernel` to your `requirements.txt` file. You can install IPykernel manually as follows:
+Positron requires the IPykernel package (version 6.19.1 or higher) to communicate with Python. IPykernel is bundled with Positron for the CPython implementation, which is the default and most widely used Python implementation.
```
isabelizimm, line 41:
```suggestion
Positron requires the [IPykernel package](https://pypi.org/project/ipykernel/) (version 6.19.1 or higher) to communicate with Python. IPykernel is bundled with Positron for the CPython implementation, which is the default and most widely used Python implementation.
```

start.qmd
```diff
@@ -38,14 +38,14 @@ rig system rtools add
...
+
+If you are using an implementation other than CPython, you'll need to install IPykernel manually in your Python environment. Positron prompts you to install IPykernel if it isn't found when you initially try to use a Python environment. Alternatively, you can add `ipykernel` to your package's existing requirements file, or install IPykernel manually as follows:
```
austin3dickey, line 43: Very small nit: it looks like the Jupyter team either [refers](https://ipython.readthedocs.io/en/stable/install/kernel_install.html) to the package as "the IPython kernel" (in prose) or `ipykernel` (when referring to the Python library). Maybe we should change the styling here to align with that by switching out the use of "IPykernel"?

austin3dickey (approved):
Thanks for the update!

start.qmd
juliasilge: I think we should add this here, but I would suggest wording like:

> If you would prefer to not have Positron's bundled ipykernel on your Python `PATH`, toggle the  `"python.useBundledIpykernel"` setting to false and instead install it manually.

juliasilge (commented):
Thank you for adding this!

isabelizimm (approved):
looks great! thank you for adding this 💯

austin3dickey (commented):
Looks great!

PR #51, opened 2025-03-04
Add download button

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/5734

This PR also clarifies the callout box that we have on <https://positron.posit.co/download> to make it a bit more prominent and clear, after feedback that it's still a bit too easy to misunderstand.


timtmok (approved):
LGTM

PR #57, opened 2025-03-12
Add some documentation around extending the connnections pane

dfalbel:
Adresses: https://github.com/posit-dev/positron/issues/6189

### DIFF
--- a/.github/styles/config/vocabularies/posit-docs/accept.txt
+++ b/.github/styles/config/vocabularies/posit-docs/accept.txt
@@ -26,3 +26,4 @@ Data Explorer
 Project Manager
 Windows
 File Explorer
+Connections Pane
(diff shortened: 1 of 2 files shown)


juliasilge (changes requested):
Thank you so much for this PR, @dfalbel!

I am a bit worried about using permalinks like this to GitHub, because it will very shortly be taking to folks to a version that may be super duper stale really quickly, depending on what else is going on in this file. What about if instead, we used an approach like this for all the GH links in these docs:


````
The Connections Pane can be extended by implementing a subclass of the 
`Connection` object. Find the definition of the `Connection` object 
[in Positron's Python extension](https://github.com/posit-dev/positron/blob/main/extensions/positron-python/python_files/posit/positron/connections.py) 
and submit a PR to the Positron repository to add your subclass.
````

Or do you maybe have a different idea?

dfalbel:
Thanks @juliasilge !

This makes sense to me. I have updated the links in the PR to point to main directly - and now mainly point to files instead of line definitions, which are likely to change.

juliasilge (approved):
Thank you so much! 🙌

PR #58, opened 2025-03-12
Update Linux dependencies

jmcphers:
We've done work recently to extend our Linux compatibility to older distributions. This change updates the download table so that it matches our new compatibility baseline.


juliasilge (approved):
Looks good!

https://deploy-preview-58--positron-posit-co.netlify.app/download.html

PR #60, opened 2025-03-23
Update Server Download Url Template

kv9898:
Since we have changed the template in #6642, the existing screenshot is misleading and should be updated with the latest representation,.


juliasilge (approved):
Thanks so much, @kv9898!

PR #66, opened 2025-04-22
Clarify that "proposed APIs" are available to all extensions

atheriel:
Positron enables so-called "proposed APIs" for all extensions by default, which is a major piece of good news for some extension authors. This commit ensures we mention this prominently in the documentation.

Addresses https://github.com/posit-dev/positron/issues/7306.


jmcphers (approved):
LGTM!

PR #71, opened 2025-05-18
Add new GH action for checking URLs

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/7546


juliasilge:
You can see what a failure looks like here (for the base URL we use to build up the download links):

https://github.com/posit-dev/positron-website/actions/runs/15094880127/job/42428262210

petetronic (approved):

PR #72, opened 2025-05-27
add fontawesome gear to Positron settings links

isabelizimm:
Addresses https://github.com/posit-dev/positron/issues/5443

I've added some custom css to add a gear in front of all links that start with `positron://settings`, which will open up settings in a Positron application. I did add the [fontawesome](https://github.com/quarto-ext/fontawesome/tree/main) extension (created and maintained by the Posit Quarto team) to make this happen, which is most of these additions 😅 

You'll be able to see an example of these gears on the `Guides > Discovering R Installations` page.

Screenshot 2025-05-27 at 3 40 17 PM


styles.css
```diff
@@ -1 +1,15 @@
 /* css styles */
+@font-face {
+  font-family: 'Font Awesome 6 Free';
+  font-style: normal;
+  font-weight: 900;
+  font-display: block;
+  src: url("_extensions/quarto-ext/fontawesome/assets/webfonts/fa-solid-900.woff2");
+}
+
+/* Add gear icon before links with href starting with "positron://settings" */
```
isabelizimm, line 10: This file is the main change, where we load font awesome and then apply the ⚙️ for any href starting with `"positron://settings"`

juliasilge (approved):
That looks so nice and works great! 🙌

PR #80, opened 2025-06-02
Add checksums to download table

timtmok:
Address https://github.com/posit-dev/positron/issues/7317

One part of the changes to show the sha-256 hashes. This only displays if the checksum file is available for the release. Clicking on the checksum copies the full checksum to the clipboard.

image

The checksum file is generated during the upload to s3 build stage when the release metadata is generated.

After this change is on the website, I can add the checksum file manually for 2025.06.0-167 and the checksums will appear.


timtmok:
I've added the checksums file to the prereleases and updated the freeze file. The preview now shows the checksums in the downloads table.

juliasilge (approved):
Thank you so much! 🙌

PR #81, opened 2025-06-03
Improve dark mode support for download table

dhruvisompura:
Addresses https://github.com/posit-dev/positron/issues/7918

We have some custom styling that makes the downloads table look disabled, however the colors used weren't compatible with dark mode and made the contents of the table illegible. I've updated the styling to rely on the body background color instead so the contents in the table are legible regardless of color theme.

https://github.com/user-attachments/assets/e6df3196-07ec-4665-bbcc-0fba4f3f8d92

### DIFF
--- a/_extensions/posit-dev/posit-docs/_extension.yml
+++ b/_extensions/posit-dev/posit-docs/_extension.yml
@@ -28,7 +28,11 @@ contributes:
       html:
         theme:
           light: [theme.scss]
-          dark: [theme-dark.scss]
+          # We are ignoring the dark theme for now because it doesn't have great support. 
+          # The work to create an official dark theme is tracked in the following issue: 
+          # https://github.com/posit-dev/product-doc-theme/issues/99. 
+          # We can uncomment the line below once the dark theme is officially supported.
+          # dark: [theme-dark.scss]
         link-external-icon: true
         link-external-newwindow: true
         toc: true


timtmok:
I can't remember if I had tried this before when adding the checkbox but it might be more obvious if the checkbox and agreement are overlaid on the table. Checking the box can make the entire overlay disappear to reveal/allow interaction with the table.

dhruvisompura:
@AshleyHenry15 I completely agree with you about the callout with a checkbox also being easy to skip right past. I do think the overall experience could be better. 

> I can't remember if I had tried this before when adding the checkbox but it might be more obvious if the checkbox and agreement are overlaid on the table. Checking the box can make the entire overlay disappear to reveal/allow interaction with the table.

@timtmok I don't think I remember seeing an overlay attempt in the PR that added the checkbox. This solution is pretty similar to the show/hide option we did try and we moved away from that experience.

I know we are trying to be GDPR compliant with the download page but I wonder if we really need the additional checkbox? I have used other desktop apps that provide a statement like: `By downloading and using Positron, you agree to the [license agreement]() and [privacy policy]().` alongside the download button/links. The act of downloading the app means you have agreed to opt-in. 

I want to suggest the above option because it would be nice if we could make the download process frictionless with the fewest clicks possible and is the cheapest option to support/maintain if this is GDPR compliant. 

I'm hoping to keep the changes cheap and minimal in this PR and just get us unblocked!

dhruvisompura:
> @dhruvisompura - just in case you want to take a look at some of the elements that we are playing around with: https://ashley.quarto.pub/theme-refresh/
> 
> very much a WIP

I'd love to take a look at this! Is there a good avenue for leaving feedback on this WIP?

AshleyHenry15:
> > @dhruvisompura - just in case you want to take a look at some of the elements that we are playing around with: https://ashley.quarto.pub/theme-refresh/
> > very much a WIP
> 
> @AshleyHenry15 I'd love to take a look at this! Is there a good avenue for leaving feedback on this WIP?

https://github.com/AshleyHenry15/theme-refresh < it is public - feel free!

The background color and some of the other elements are not set in stone - Greg S and I have been messing around with colors and what not in our "spare time". 

Posit blue is not super friendly for dark themes, unfortunately. So, we end up having to adjust it quite a bit.

I stumbled upon https://tailwindcss.com/docs/installation/using-vite docs and I like their background color so I was going to give that a try and test Posit blue options again that.

AshleyHenry15:
@dhruvisompura - this is the repo for the current theme extension as well (if you want to open issues for what is currently deployed): https://github.com/posit-dev/product-doc-theme/

PR #82, opened 2025-06-03
Add additional commit that was missed

jthomasmock:
Missed one commit 😢


juliasilge:
I don't think this version of how to enable the setting is super clear; I'll do one more PR.

PR #86, opened 2025-06-06
Release notes for RC

juliasilge:
No rush on this, but I want to start getting feedback on our initial go at release notes for RC and beyond. Some things to note:

- This new file `release.md` can be slurped up into the product before a release is built, for in-product release notes.
- We could use a strategy like `next.md`, `release.md`, and then, say, `release-2025-07.md`.
- We could work on PRs to `next.md` during the release, to prep the release notes.
- We would need to finalize the release notes _before_ a release is made but after a branch is cut.
- At a release, `next.md` can move to `release.md` and the current `release.md` can move to something like `release-2025-07.md`.
- We could keep all the old release notes in the directory together, and make a page where they are available for perusing, linked off from the downloads page.


jonvanausdeln:
This looks good to me.  I do like keeping old release notes around, I know users will ask about them and they are handy for QA and issue bisecting.  The use of `next.md` seems reasonable to me.  We should think about some GHActions to help automate that transition.

timtmok:
@juliasilge It's just searching for an element to determine where to put the checkbox. It's a quick change in Positron to use the first `h3` element for placing the checkbox. I like it more if the release notes were in one file and we can display it where we need it. Then nothing needs to change here. It would look like this:

image

Or placing any kind of element that can be found with a HTML document selector would work too.

@petetronic What do you think about placing the release notes in S3? They'll be available through the CDN and that would be the link to use. We have the credentials set up to invalidate the file if we need to update the release notes. Positron itself can build the URL to it using the base update URL (https://cdn.posit.co/positron) from the `product.json`. This can all be automated in the release process so that the installers and release notes are published at the same time.

juliasilge:
@timtmok RE that checkbox:

> Or placing any kind of element that can be found with a HTML document selector would work too.

How hard would it be to show up _after_ the highlights section, i.e. before the second H3 header?

timtmok:
> @timtmok RE that checkbox:
> 
> > Or placing any kind of element that can be found with a HTML document selector would work too.
> 
> How hard would it be to show up _after_ the highlights section, i.e. before the second H3 header?

@juliasilge It wouldn't be too hard. I've been thinking it would be best to use an HTML element place the checkbox exactly where we want and be able to see in the markdown where it would go.

timtmok:
I've been thinking about how to publish the release notes as an inclusion in the downloads page and as a separate markdown file. Publishing to the CDN is a bit more complicated with adding another GH action to upload to S3. If we include the file as a resource in the Quarto website, it's a much simpler change.

Adding this to `_quarto.yml` would get what I need to show the release notes in Positron:

```yaml
resources:
  - "release-notes/release.md"
```

release-notes/release.md
```diff
@@ -0,0 +1,56 @@
+### Highlights
```
timtmok, line 1: After trying out some release note changes in Positron, it would be nice to have the filename always tied to the release. So `release-2025.07.0.md` would be preferred. This allows past releases to always fetch the release notes for its version.

It could also be just the hosted Netlify markdown file that gets this filename.

timtmok: I think this might be the way to go. I'll add a step to the GH action in this repo that will upload the `release.md` as `release-{Positron version}.md`. I'll work with Pete to ensure the correct AWS credentials are added.

timtmok:
@juliasilge I've updated the Github action to publish the release note. I'll wait for this to merge before merging mine: https://github.com/posit-dev/positron-website/pull/97

juliasilge:
OK, I think this is the last of a flurry of PRs for our FIRST STABLE RELEASE! 🥳

jonvanausdeln (approved):
LGTM!

PR #88, opened 2025-06-09
Add info on daily builds to docs

juliasilge:
We have invited external folks to use the dailies and it seems to be going well: https://github.com/posit-dev/positron/discussions/7957

Once we have this documented, we can deprecate our weekly GH releases, either right away or at the end of this prerelease milestone.


juliasilge:
I decided to leave the info on Windows package managers here for now?

jonvanausdeln (approved):
LGTM!  I'm not huge fan of the scoop and winget since we don't maintain those and can't guarantee it's validity.. but I see the value and understand the community desire for it.

PR #89, opened 2025-06-12
Use more fancy links for Positron settings

juliasilge:
I noticed a few of these, so I did a search to try to find all of them.


isabelizimm (approved):
some nits about adding `setting` to the link text, but this looks good!

PR #96, opened 2025-06-24
Adding workbench information into its own section

AshleyHenry15:
As suggested in issue [#7538](https://github.com/posit-dev/positron/issues/7538), I added a standalone workbench section.

It seemed awkward to name the section something "workbench" related so I tried to be a bit more vague and focusing on user need / how they may search the docs or google.

I am getting a feel for these docs so please feel free to let me know if I went about this all wrong and how you would like the content to be presented.

Note: The issue mentioned potentially adding workbench-specific items like Jobs and Managed Creds, I did not cover this in the PR. It seemed like maybe that it's something we may want to cover in the Workbench guide instead?


AshleyHenry15:
@melissa-barca tagging you as a reviewer since you had added some input to the original issue. Please feel free to remove yourself if you don't have the bandwidth.

PR #98, opened 2025-06-25
first draft of goals

isabelizimm:


about.qmd
```diff
@@ -0,0 +1,15 @@
...
+title: "About Positron"
+---
+
+The goal of Positron is to create an IDE for data science that supports the full scope of modern scientific and data-driven work, across languages, tools, and workflows. We aim to do this by:
+
+-   Providing a cohesive and thoughtful experience for writing code, performing analyses, and exploring data. Whether you’re working in Python, R, or both, Positron combines notebooks, scripts, and consoles with native support for plotting and data output.
+
+-   Supporting the kinds of workflows that are central to data science. This includes iterative development, rich interactive output, and publication-quality documents. Positron provides built-in support for notebooks, Quarto, web applications, and more, enabling a range of common authoring and analysis tasks.
+
+-   Offering a modern, extensible editor built on open standards. Positron is built on the same foundation as Visual Studio Code, with custom features and UI elements designed specifically for data science. This allows for a wide array of extensions and customization—while also enabling deeper integration of language-specific tools.
```
jthomasmock, line 11: Offering a modern, extensible editor built on open standards. Positron is built on Code OSS, the same foundation as Visual Studio Code, with custom features and UI elements designed specifically for data science. This includes support for a wide array of VSIX extensions and customization — while also enabling deeper integration of language-specific tools.

jthomasmock (approved):

about.qmd
juliasilge: The only other thing I would say here is to avoid the em dash maybe? I have never been a fan 😆 and now apparently people think it means AI wrote the text.

PR #109, opened 2025-07-02
Use release version in footer now

juliasilge:
This PR removes the beta footer and uses whatever the current release version is.


austin3dickey (approved):
Thank you!!

PR #110, opened 2025-07-02
fix download CDN url on Getting Started page

sharon-wang:
On the Getting Started page (https://positron.posit.co/start.html#install-positron), the old release url is being used for the download button:

image

The issue is that the url is still set to the prereleases CDN url for that page. The url has now been updated to the releases CDN url. For Mac, `-universal` has been appended to the filename before the `.dmg` extension, as the universal filename has changed since we started building arch-specific binaries.

---

I briefly explored moving the CDN url to an env var in `_environment`; however, when passing the url from the quarto file to the `_download.html` file, it ends up as `undefined`, probably due to some issues with escaping the string. So, an improvement for another day unless anyone has ideas!

---

preview url: https://deploy-preview-110--positron-posit-co.netlify.app/start.html#install-positron


juliasilge:
It is `"Positron-{release_version}-universal.dmg"` now.

PR #111, opened 2025-07-06
Try out using `kbd` shortcut

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/8238

I tried out using the Quarto `kbd` shortcut here in this PR: https://quarto.org/docs/authoring/markdown-basics.html#keyboard-shortcuts

There are two problems I am observing here:

- The first is basically the same as https://github.com/quarto-dev/quarto-cli/issues/5879 where we can notice that these don't look so great in general
- The second is that there is something wrong with how Alt is being rendered with the separate `-`; I can't seem to get it to work correctly

For example, if we have something like this:

```
{{< kbd mac=Command-Alt-M win=Ctrl-Alt-M linux=Ctrl-Alt-M >}}
```

It renders like this on my Mac right now:

Screenshot 2025-07-06 at 3 56 09 PM


juliasilge:
Many thanks to the Quarto team for their direction on Alt vs. Option on Windows vs. Mac!

New preleases of Quarto now include fixes for how the `kbd` shortcode works, but I think maybe we can just live with the current, slightly ugly state until the next Quarto release (rather than move our whole site to prelease Quarto).

jennybc:
One comment re: whether we want this: R used to (or maybe still does?) offer different help topics or help topic content depending on the running OS (Windows vs. not Windows). And this practice has been eliminated because it's just confusing for folks who use 1 OS and need to communicate with or teach folks who use another OS.

Likewise, in the tidyverse we pretty much have an unofficial (or maybe official) stance on *never* doing this in docs.

So there is some argument for allowing mac folks to see Windows shortcuts and vice versa. I do understand there are good arguments in favor of trying to show the OS-specific shortcut. I just wanted to put this other consideration into the mix.

Somehow I feel like in-product displays of shortcuts are different from the display in the official documentation. Customizing the former feels right but it's less clear cut for the latter.

jennybc:
The hover feature the VS Code uses to display the shortcuts for the other OSes is pretty nifty:

Screenshot 2025-07-10 at 10 34 48 AM

juliasilge:
Oh yeah, that is great! I opened a feature request for the Quarto `kbd` shortcode here: https://github.com/quarto-dev/quarto-cli/discussions/13058

juliasilge:
I think this is ready for some review now and a decision on whether we want to do this.

- Check out a keyboard shortcut heavy page here: https://deploy-preview-111--positron-posit-co.netlify.app/keyboard-shortcuts.html
- Check out how it looks when in text here: https://deploy-preview-111--positron-posit-co.netlify.app/data-explorer.html#grid-selection-and-copy-and-paste

Thanks for sharing some reason why we might _not_ want to do this, @jennybc. I'm going to ping both @jthomasmock and @petetronic for some review. 

I'll highlight that Tom is out until Monday, when I will then be out. I am totally fine with someone else making any final changes or merging this next week while I am out.

juliasilge:
I updated to the latest Quarto prerelease that has many of the latest `kbd` improvements, and I think we _should_ go ahead and use that for rendering our site. It includes those tooltips that show the keyboard shortcuts for the other OSes as well:

Screenshot 2025-07-10 at 4 14 07 PM

petetronic (approved):

PR #95, opened 2025-06-23
user interface page

rodrigosf672:
Partially addresses #7541 by adding docs for the command palette.

Based on initial feedback from @jthomasmock and @juliasilge, I have made some changes and put together as a `.qmd` file to suit the requirements for documentations. It's my first time opening a PR in this repo, so please reach out in case I'm missing any standards/processes. Thanks!

PR #113, opened 2025-07-07
Update user-interface.qmd

ryjohnson09:
The project is in the top bar, not the title bar.


juliasilge:
We're going to get this addressed in #95. Thank you!

PR #118, opened 2025-07-11
Command palette linkage

dhruvisompura:
Addresses [#7541](https://github.com/posit-dev/positron/issues/7541)

Now that we have a doc for the command palette, this PR :
- updates the VS Code migration guide doc to link to the command palette guide
- Removes the inline command palette shortcut we mention everywhere else. I think we can rely on the user reading the command palette guide to understand how to bring it up.
- Updates capitalization and wording in some places that mention the command palette to be consistent in how we talk about using the command palette.

### DIFF
--- a/action-bars.qmd
+++ b/action-bars.qmd
@@ -43,4 +43,4 @@ The Top Action Bar helps you manage your work. On the left, you will find action
 
 ## Wrapping up
 
-Action Bars surface the most important and frequently used actions in each area of the product, so you do not have to rely on the command palette to find everything. Because they are contextual, Action Bars present the tools you need right where you need them, helping you stay focused and work more efficiently in Positron.
+Action Bars surface the most important and frequently used actions in each area of the product, so you do not have to rely on the Command Palette to find everything. Because they are contextual, Action Bars present the tools you need right where you need them, helping you stay focused and work more efficiently in Positron.
--- a/add-to-path.qmd
+++ b/add-to-path.qmd
@@ -11,7 +11,7 @@ See the instructions below on how to add Positron to your PATH based on operatin
 
 ### macOS
 
-After you have installed Positron, you can add it to your PATH by running the _Shell Command: Install 'positron' command in PATH_ command from the Command Palette.
+After you have installed Positron, you can add it to your PATH via the Command Palette. Press {{< kbd mac=Command-Shift-P win=Ctrl-Shift-P linux=Ctrl-Shift-P >}} to open the palette and type *Install 'positron' command in PATH*.
 
 ![Add Positron to your PATH](images/positron-path.png)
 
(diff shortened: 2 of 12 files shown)


jennybc:
I love that we are standardizing this, it will be a big relief.

But yeah I do also feel this goes to far in terms of DRYing things out, which often works better for code than for docs. Meaning the overall lack of link and lack of keyboard shortcut.

Thinking about how lovely pkgdown and autolinking works over in the R world ... how feasible is it for us to use some sort of :sparkles: Quarto magic :sparkles: where we could type "Command Palette" in a certain way (in square brackets or what have you), which would then apply our chosen treatment (autolinking to the main guide? showing OS-appropriate keyboard shortcut on hover)? That would be an up-front investment but then trying on different treatments because relatively painless and any choice we make gets applied everywhere.

Is that what [Quarto shortcodes](https://quarto.org/docs/extensions/shortcodes.html) are for?

dhruvisompura:
> Thank you so much for this thorough work! 🏆 
> 
> 
> 
> I have some mixed feelings about removing _all_ the places we tell people how to bring up the command palette, but I acknowledge that may be my RStudio bias showing.
> 
> 
> 
> I do notice that in the VS Code docs, they show the keyboard shortcut basically every time they talk about the command palette. See for example: https://code.visualstudio.com/docs/sourcecontrol/overview
> 
> 
> 
> What do you think, @jennybc? Do you feel like we have this too liberally sprinkled now?

I have no qualms with showing the keyboard shortcut.

We don't mention the keyboard shortcut every time we talk about the command palette. It was a small change for consistency but I don't think we have to go hard in one direction or the either. I can undo the change and leave the keyboard shortcuts in!

jennybc:
@dhruvisompura Looks like we were typing at the same time!

dhruvisompura:
> I love that we are standardizing this, it will be a big relief.
> 
> But yeah I do also feel this goes to far in terms of DRYing things out, which often works better for code than for docs. Meaning the overall lack of link and lack of keyboard shortcut.
> 
> Thinking about how lovely pkgdown and autolinking works over in the R world ... how feasible is it for us to use some sort of ✨ Quarto magic ✨ where we could type "Command Palette" in a certain way (in square brackets or what have you), which would then apply our chosen treatment (autolinking to the main guide? showing OS-appropriate keyboard shortcut on hover)? That would be an up-front investment but then trying on different treatments because relatively painless and any choice we make gets applied everywhere.
> 
> Is that what [Quarto shortcodes](https://quarto.org/docs/extensions/shortcodes.html) are for?

I've reverted my change, and left all the keyboard shortcuts in. I think a shortcode could work - I can look into this in a follow up PR!

jennybc (approved):
IIUC correctly this is not the final state, but it's a move in the right direction (?). More consistency around "Command Palette" and correcting a hyperlink.

And we'll continue to iterate on a more definitive solution, possibly using a Quarto short code.

PR #122, opened 2025-07-15
Change "Run Interactive Apps" to "Develop Data Apps"

nrohr:
I propose updating some of the copy about running interactive apps to specify "data apps", which is becoming industry standard terminology, e.g.:
- Dash: "Dash is the original low-code framework for rapidly building data apps in Python."
- Streamlit: "A faster way to build and share data apps"
- Databricks Apps: "Quickly build secure data and AI apps"

I personally find myself searching for "data apps" on the documentation website and having to remind myself we currently say "interactive apps". I believe this change would help more users find this section of our docs.

### DIFF
--- a/develop-data-apps.qmd
+++ b/develop-data-apps.qmd
@@ -1,8 +1,8 @@
 ---
-title: "Run Interactive Apps"
+title: "Develop Data Apps"
 ---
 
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
Thanks for this contribution, @nrohr! Great feedback 🚀

PR #125, opened 2025-07-22
Add some guidance for completions with remote SSH sessions

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/8409 in that it at least provides some documentation, but we will not close that issue. Instead, we'll move it back to backlog and we'll eventually improve the experience in the product itself.

### DIFF
--- a/assistant.qmd
+++ b/assistant.qmd
@@ -42,10 +42,12 @@ Copy and save the API key to a password manager or another secure location. In o
 At this time, only **GitHub Copilot** is available for inline code completions.
 :::
 
-If you have a GitHub account with Copilot enabled, you can authenticate to GitHub Copilot in Positron Assistant in the language model dialog. The authentication flow will open up a browser and direct you to sign into GitHub. Once authenticated, Copilot will be available for inline code completions.[^1]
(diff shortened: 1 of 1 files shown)


assistant.qmd
```diff
@@ -42,7 +42,10 @@ Copy and save the API key to a password manager or another secure location. In o
 At this time, only **GitHub Copilot** is available for inline code completions.
 :::
 
-If you have a GitHub account with Copilot enabled, you can authenticate to GitHub Copilot in Positron Assistant in the language model dialog. The authentication flow will open up a browser and direct you to sign into GitHub. Once authenticated, Copilot will be available for inline code completions.
+If you have a GitHub account with Copilot enabled, you can authenticate to GitHub Copilot in Positron Assistant in the language model dialog. The authentication flow will open up a browser and direct you to sign into GitHub. Once authenticated, Copilot will be available for inline code completions.[^1]
+
+
+[^1]: If you are using Positron with a [remote SSH](remote-ssh.qmd) session, you will need to authenticate to GitHub on the _remote server_.
```
juliasilge, line 48: Is a footnote enough? Or not prominent enough? We've already had ~3 reports of folks running into this.

sharon-wang: Would we want to link to the issue as part of this note?

juliasilge: Yeah, probably helpful for folks

sharon-wang (approved):
Thank you for clearing this up in the docs! 🙌

PR #132, opened 2025-07-25
Update page on license: add a title, add link to air

juliasilge:
I noticed that this page currently doesn't have a nice title, which looks bad in browsers: https://positron.posit.co/licensing.html

I also added a link to air.


jthomasmock (approved):
LGTM - thank you!

PR #138, opened 2025-07-31
Remove unused GH action

juliasilge:
I have realized that this GH action doesn't do anything. 🙈 

We render this _on Netlify_ as controlled here:

https://github.com/posit-dev/positron-website/blob/3a127d0f9d664c56208ed938b6cd87fd69b1fc88/netlify.toml#L2

What made me realize that this is not doing anything is that this action failed for this PR (because that PR requires using a Quarto prerelease, and this action just uses the regular Quarto release):

https://github.com/posit-dev/positron-website/actions/runs/16634976637

But notice that the site is published just fine, with the new theme:

https://positron.posit.co/


jonvanausdeln (approved):
LGTM!

PR #140, opened 2025-07-31
Switch to the "neutral" mermaid theme temporarily

juliasilge:
After we applied the new theme, the mermaid diagrams do not render well:

Screenshot 2025-07-30 at 6 57 43 PM

Just until we can get this straightened out, let's go with the dead boring "neutral" theme, which is at least readable. 😅

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -98,4 +98,6 @@ website:
 format:
   html:
     css: styles.css
+    mermaid: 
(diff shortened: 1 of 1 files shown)


juliasilge:
See the now-extremely-gray diagram here: https://deploy-preview-140--positron-posit-co.netlify.app/remote-ssh.html

PR #144, opened 2025-08-04
Adjusting padding for keyboard shortcuts

AshleyHenry15:
Slightly decreases the padding for the keyboard styles.

```
kbd {
    padding: .2rem .2rem;
}
```

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -30,7 +30,7 @@ nav .nav-item:not(.compact) {
 
 /* Keyboard shortcuts */
 kbd {
-    padding: .2rem .2rem;
+    padding: .2rem .4rem;
 }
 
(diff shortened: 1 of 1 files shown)


AshleyHenry15:
Hey @juliasilge - I committed your suggestion but I think that I will have to have you test it because I am not experiencing the same behavior or squishing on my end :(

AshleyHenry15:
I am going to circle back to marketing to let them know that the font type and size requires a larger line height and that it may slightly differ from the guidelines provided.

juliasilge:
Awesome! Thank you for this; we'll look to see where we need to head for consistency with brand guidelines in the longer run, but for now I'm going to merge this so we have more readable keyboard shortcuts.

PR #145, opened 2025-08-04
Release 2025.08

juliasilge:
This PR updates the website for the latest monthly release 2025.08.0-130.


_freeze/start/execute-results/html.json
juliasilge, line 1: We also need to render and check in `_freeze` for `/start` now that we have a button there to download as well.

release-notes.qmd
```diff
@@ -0,0 +1,7 @@
+---
+title: "Older Releases"
+---
+
+## 2025.07
```
juliasilge, line 5: Once we have more than one thing here, I'll make a table instead of just listing this out in one giant page.

PR #149, opened 2025-08-11
Update Connections Pane docs

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/7556

We can't make new screenshots right now because of https://github.com/posit-dev/positron/issues/8964 😬


juliasilge:
@dfalbel Do you have any ideas for a troubleshooting section? Do we have enough info gathered anywhere to add a section on that right now? We may need to wait on that for further data on what kinds of problems people actually have.

dfalbel (approved):
LGTM!

PR #162, opened 2025-08-29
Clarify keybinding in release notes

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/9213

Since we will keep these around for the future, let's go ahead and fix this. I do _not_ think we need to update the release notes in the S3 bucket since we are about to publish new ones.


jonvanausdeln (approved):
LGTM!

PR #168, opened 2025-09-05
Feat/homepage redesign part 3

cindyytong:


cindyytong:
For conf will make a slimmed down banner in a separate PR.  Will revisit this after conf

PR #172, opened 2025-09-10
Move button back to left, temporarily

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/9374 as a temporary stopgap, until we can figure out top bar CSS more broadly.


juliasilge:
We're all back to aligned with this change:

Screenshot 2025-09-10 at 10 42 50 AM

I'm going to merge this in, and then we can take our time on fixing in the long term.

PR #178, opened 2025-09-18
fix url ref due to nesting styles.css under css/

cindyytong:
See [thread](https://positpbc.slack.com/archives/C04FPQK3H9C/p1758201865861269)

Because I moved styles.css under css/, I need to update any url references within the styles.css sheet.  This is causing icons not to load .

Local changes show the icon appearing:
Screenshot 2025-09-18 at 2 46 57 PM

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -4,7 +4,7 @@
   font-style: normal;
   font-weight: 900;
   font-display: block;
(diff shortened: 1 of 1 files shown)


petetronic (approved):
Tested preview and icon was fixed

juliasilge (approved):
Thank you!

PR #179, opened 2025-09-22
Add a little section on retention policy

juliasilge:
I think it's best that we not muddy this with any info about the pre-releases.


updating.qmd
```diff
@@ -20,6 +20,10 @@ If you change the setting to `"dailies"`, what can you expect?
 - These daily builds only go through automated testing before being made available, not our full QA process. If you consider yourself an adventurous early adopter, this may be a good fit for you!
 - If some daily build happens to be entirely broken for you, the best step for you to take is to close Positron and [install the latest monthly release](download.qmd).
 
+## Retention policy
```
petetronic, line 23: Should we put this as ### level under daily builds?

petetronic: Should we use current setting of 60 days?

petetronic (approved):

PR #180, opened 2025-09-22
Support uv for managing Python

austin3dickey:
This is an often-requested change to let folks know that `uv` is a good choice for managing Python versions for Positron.


isabelizimm (approved):
looks good! At some point, I wonder if we want to have a "how to choose which is right for you" piece, to help folks make the `pyenv` vs `uv` decision. But I think it's a good start to mention that WE ❤️ UV!

PR #181, opened 2025-09-24
Adjust organization for interpreter pages

juliasilge:
Addresses #7946

We will follow up with:

- https://github.com/posit-dev/positron/issues/9404
- https://github.com/posit-dev/positron/issues/7208


cindyytong (approved):
LGTM Ty

PR #183, opened 2025-09-30
Polish remote SSH docs

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/7554

These docs are in very good shape, but I added just a few points based on questions we've had in recent months.


austin3dickey (approved):
lovely, thanks!

PR #185, opened 2025-10-02
An attempt to theme the mermaid diagrams

juliasilge:
@AshleyHenry15 I played around with this a bit tonight and came up with this, which does work. What is good about this option is:

- the text is still big enough relative to the rest of the page
- we can read the text against the background of the diagram shapes

What do you think about controlling the mermaid diagram look this way? Could we use this type of CSS in the theme itself, or is it better to take this approach?

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -127,6 +127,4 @@ website:
 format:
   html:
     css: css/styles.css
-    mermaid: 
-      theme: neutral
     link-external-filter: '^(?:http:|https:)\/\/(?:positron|cdn)\.posit\.co(?:\/|$)|^(?:http:|https:)\/\/.*--positron-posit-co\.netlify\.app(?:\/|$)|^(?!https?:\/\/)'
--- a/css/styles.css
+++ b/css/styles.css
@@ -50,6 +50,13 @@ kbd {
   margin-right: 5px;
 }
 
+/* Mermaid diagram customization */
+:root {
+  --mermaid-label-fg-color: #3a78b1; /* positron-blue for text color */
+  --mermaid-node-bg-color: #e6eef6;  /* Light blue to go with positron-blue */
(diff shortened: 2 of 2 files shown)


juliasilge:
You can check out what it looks like here:

https://deploy-preview-185--positron-posit-co.netlify.app/remote-ssh.html

AshleyHenry15:
I could add this to the positron "theme" branch but this isn't currently passing accessibility checks as is. 

Also, a few other failures have popped up, it seems that "code" used in URLs/links fails as well. That isn't a use case that I had tested previously.

AshleyHenry15:
@juliasilge these styles pass accessibility and also rounds some of the edges so things look a bit nicer. 

I can add these to the "theme" to bring in later but you can try to add these locally to see if they work?

I had to go this route because the styles you had didn't cover "all" types of diagrams (I will attach an image so you can se what I mean)

```

/* Target all Mermaid diagrams */
[id^="mermaid-"] .cluster rect {
  fill: #f9f9fb !important;
  stroke: #17212b !important; // posit-dark-blue-3
}

[id^="mermaid-"] .node rect {
  fill: #e6eef6 !important;
  stroke: #17212b !important; // posit-dark-blue-3
}

[id^="mermaid-"] .nodeLabel {
  fill: #17212b !important; // posit-dark-blue-3
}

[id^="mermaid-"] .edgeLabel p {
  color: #17212b !important; // posit-dark-blue-3
}

[id^="mermaid-"] foreignObject:has(.edgeLabel) {
    background-color: #f9f9fb !important;
    border-radius: 8px;
}

[id^="mermaid-"] p {
    margin-inline: auto !important;
    max-width: max-content;
    border-radius: 8px;
    padding: 0 0.75rem;
}

[id^=mermaid-] .labelBkg {
    background-color: transparent !important;
}
```

image

PR #190, opened 2025-10-04
Migrate videos to non-tracking option with no cookies

juliasilge:
First attempt here is to use videos in the repo itself, tracked via git lfs so we don't get annoying slow behavior in the repo.


juliasilge:
Seems pretty good to me! 

https://deploy-preview-190--positron-posit-co.netlify.app/

PR #193, opened 2025-10-06
Update Positron Pro info on `/download`

juliasilge:
Positron Pro is generally available on WB as of 2025.09.0.


juliasilge:
Updated language is visible at https://deploy-preview-193--positron-posit-co.netlify.app/download.html#positron-pro-on-posit-workbench

PR #196, opened 2025-10-07
Add text of license to website, for linkability

juliasilge:
Addresses https://github.com/posit-dev/positron-builds/issues/540 by putting the text of the license here, so we can link to it.

I styled the HTML mostly the same as https://www.elastic.co/licensing/elastic-license, but am not strongly advocating for that if there are other considerations.


juliasilge:
Preview is here, for any thoughts on the styling: https://deploy-preview-196--positron-posit-co.netlify.app/licensing.html

PR #200, opened 2025-10-08
Support Python 3.14

austin3dickey:


juliasilge (approved):
Thank you! 🙌 

Am I correct in understanding that we should merge this once the release is out? In another day or so?

PR #324, opened 2026-04-06
Add language server section to Python guide

austin3dickey:
Here we add an explicit section in the Python guide about the language server. Most of this is copied from the blog post.


cindyytong (dismissed):
LGTM

isabelizimm (dismissed):

PR #331, opened 2026-04-13
Document pixi support

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/12249

I mentioned pixi in two places:

- in our list of supported Python managers, which gets re-used several places
- on the R installations page, under "Customizing R discovery"

### DIFF
--- a/_python-managers.qmd
+++ b/_python-managers.qmd
@@ -1,4 +1,4 @@
 * **venv**: Standard library virtual environments created with `python -m venv`
 * **uv**: Virtual environments and Python installations managed by `uv`
 * **pyenv**: Python installations managed by `pyenv`, including virtual environments
-* **conda**: Conda environments created with `conda create` or `mamba create`
+* **conda**: Conda environments created with `conda create` or `mamba create`, as well as pixi environments created with `pixi init`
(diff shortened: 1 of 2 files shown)


_python-managers.qmd
```diff
@@ -1,4 +1,4 @@
 * **venv**: Standard library virtual environments created with `python -m venv`
 * **uv**: Virtual environments and Python installations managed by `uv`
 * **pyenv**: Python installations managed by `pyenv`, including virtual environments
-* **conda**: Conda environments created with `conda create` or `mamba create`
\ No newline at end of file
+* **conda**: Conda environments created with `conda create` or `mamba create`, as well as pixi environments created with `pixi init`
```
juliasilge, line 4: What do we think about this? Do we like it this way, or as a separate bullet in the list? Which would make the most sense to folks?

juliasilge (commented):
You can see one of the spots where we use that `include` here:

https://deploy-preview-331--positron-posit-co.netlify.app/migrate-vscode.html#language-support

_python-managers.qmd
isabelizimm: I think this is good as-is, in one bullet point!

isabelizimm (dismissed):

PR #332, opened 2026-04-14
Add new `workflow_call` trigger for GH action to release docs bundles

juliasilge:
Addresses https://github.com/posit-dev/positron-builds/issues/908

We need to be able to call this workflow here over from the daily release workflow, to get the automated PRs for Workbench set up correctly.

Right now I have `secrets: inherit` to pass the calling workflow's secrets to the called workflow. We need `AWS_S3_RW_ROLE`, `AWS_CDN_RW_ROLE`, and `AWS_CDN_DISTRIBUTION_ID`. I think these are repo-level secrets scoped only to `positron-website`, which means this won't work right now.

### DIFF
--- a/.github/workflows/release-docs-bundles.yml
+++ b/.github/workflows/release-docs-bundles.yml
@@ -12,6 +12,13 @@ on:
         required: true
         type: string
         default: releases
+    secrets:
+      AWS_S3_RW_ROLE:
+        required: true
+      AWS_CDN_RW_ROLE:
+        required: true
+      AWS_CDN_DISTRIBUTION_ID:
+        required: true
   workflow_dispatch:
     inputs:
       version:
--- a/.github/workflows/release-docs-bundles.yml
+++ b/.github/workflows/release-docs-bundles.yml
@@ -1,4 +1,17 @@
 on:
+  workflow_call:
+    inputs:
+      version:
+        required: false
+        type: string
+      overwrite:
+        required: false
+        type: boolean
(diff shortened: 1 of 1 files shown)


juliasilge:
OK, I think I've got an approach for the AWS creds.

When `release-docs-bundles.yml` runs via `workflow_dispatch` from `positron-website` here (a public repo), it reads `AWS_CDN_RW_ROLE` and `AWS_CDN_DISTRIBUTION_ID` as **secrets**; I think this is the right choice since vars in a public repo are visible to anyone. 

In `positron-builds` (a private repo), those same values happen to be stored as **vars** rather than secrets, alongside `AWS_S3_RW_ROLE` which is a secret in both repos. To make the cross-repo `workflow_call` work, we explicitly declare all three as secrets in the `workflow_call` trigger of `release-docs-bundles.yml`, then thread them down the call chain: `build-release.yml` reads `AWS_S3_RW_ROLE` from `secrets`, and `AWS_CDN_RW_ROLE`/`AWS_CDN_DISTRIBUTION_ID` from `vars`, passing all three as secrets to `update-workbench.yml`, which in turn passes them as secrets to `release-docs-bundles.yml`. This means the called workflow always receives them via `secrets` regardless of how the caller stores them, and the existing `workflow_dispatch` path from `positron-website` here is completely unchanged.

Does that seem reasonable?

melissa-barca (approved):
This plan makes sense to me!

PR #335, opened 2026-04-16
Add workflow to update all-releases.json after docs are published

melissa-barca:
Part 2 of posit-dev/positron-builds#934

Brings over the workflow to update all-releases.json and associated python script with a few changes:
- Workflow is now precedented by the release docs job 
- Include `docs_url` for each release entry (addressing posit-dev/positron#12197`)
- Updated deduplication logic to compare year and month rather than exact version so that only the latest patch is included for each version
- Removed fields that weren't being used: name, commit, and codeoss_verison

### DIFF
--- a/.github/workflows/release-docs-bundles.yml
+++ b/.github/workflows/release-docs-bundles.yml
@@ -48,6 +48,8 @@ jobs:
     permissions:
       contents: write
       id-token: write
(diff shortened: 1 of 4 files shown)


juliasilge (approved):
Thank you!

PR #337, opened 2026-04-20
add flag to skip writing release body during release workflow

sharon-wang:
- addresses https://github.com/posit-dev/positron-builds/issues/949

When this workflow is called from positron-builds (via `workflow_call`), the release already exists with a release body. Previously, `softprops/action-gh-release@v2` would overwrite that body with a generic "Documentation for <url>" message, which is coming from this repo's workflow.

This PR adds a `create_release` input (defaults to `true` so the workflows in this repo continue to behave as before) that controls whether the workflow creates a new release or just uploads assets to an existing one:

- `create_release: true` (default, when a release is created manually from this repo): Creates the release with body + files using softprops, same as before.
- `create_release: false` (positron-builds caller): Uploads zip files to the existing release using `gh release upload`, and does not modify the release body or name.

This does rely on the fact that the release already exists when `create_release: false` is used, but that should be the case in the positron-builds workflows.

We'll need to confirm this is working by reviewing the next positron-builds PR that calls this workflow with `create_release: false` and verifying that the release body is not overwritten.

### DIFF
--- a/.github/workflows/release-docs-bundles.yml
+++ b/.github/workflows/release-docs-bundles.yml
@@ -12,7 +12,7 @@ on:
         required: true
         type: string
         default: releases
-      set_release_body:
+      create_release:
         required: false
         type: boolean
         default: true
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
Thank you! 🚀

PR #338, opened 2026-04-20
try another linter to reduce flaky lint CI failures

sharon-wang:
the lint job fails a lot 😢, and results in having to check the lint job output, only to see that the failed URLs are actually fine, but we got blocked or rate limited in some way during the url checking.

Failure rate of 69%:
image

---

This PR switches us to lychee for the lint checker (https://github.com/lycheeverse/lychee-action), which has more customizability and seems to avoid the bot checking / rate limiting by running concurrent link checking requests and using exponential backoff for retries.

We also get a bit more info about the state of our URLs. In particular, there are quite a few redirects that we could consider updating our urls to the new url for. The github action summary shows more detail: https://github.com/posit-dev/positron-website/actions/runs/24674497576/attempts/1#summary-72154467820

The lint job still takes about the same amount of time (~1.5 mins), maybe slightly less than before on average, since the lint job sometimes took > 2 or 3 minutes, whereas this one currently is pretty consistently taking ~1.5 mins.

The goal here is not necessarily a lower failure rate, but definitely a lower failure rate for URLs that are actually working. In particular, we tend to see Microsoft and Snowflake urls fail in the lint job even though the URLs are still live, so these should stop failing!

### DIFF
--- a/lychee.toml
+++ b/lychee.toml
@@ -6,9 +6,10 @@ exclude = [
   "https://cdn.posit.co/positron/prereleases/",
   "https://positron.posit.co/images/",
   "https://github.com/posit-dev/positron/issues",
-  "positron://settings/"
+  "positron://settings/",
+  "https://forms.gle/DRCf8wgbaafz5fcG8" # testimonial Google Form (requires login)
 ]
 
-# 403 (Forbidden) and 429 (Too Many Requests) are returned by sites like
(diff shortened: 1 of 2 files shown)


lychee.toml
sharon-wang, line 1: lychee config docs: https://lychee.cli.rs/guides/config/

juliasilge (approved):
Thank you for digging into this! 🏆

PR #339, opened 2026-04-21
set repo for uploading assets to release

sharon-wang:
- a follow-up to https://github.com/posit-dev/positron-website/pull/337 to address https://github.com/posit-dev/positron-builds/issues/949
- fix `gh release upload` targeting the wrong repo when called via `workflow_call` from `positron-builds`, resulting in the release tag not being found, and the assets not getting uploaded

Passing `--repo` should ensure that the repo is the one from which the workflow is run (builds repo), rather than the repo where the workflow lives (this website repo).


juliasilge (approved):
Thanks for tracking all this down! 🙌

PR #341, opened 2026-04-23
add callout linking to posit assistant docs

sharon-wang:
- addresses https://github.com/posit-dev/positron/issues/13147

### DIFF
--- a/assistant.qmd
+++ b/assistant.qmd
@@ -36,6 +36,12 @@ listing:
         link: "https://posit.co/blog/?post_tag=ai-newsletter"
 ---
 
+::: {.callout-note}
+## Looking for Posit Assistant?
+
+Posit Assistant is a new AI coding experience currently available as preview feature. To learn how to enable it and get set up, visit the [Posit Assistant documentation](https://pos.it/assistant-docs).
+:::
+
 Positron Assistant is an AI coding assistant built specifically for data science workflows. It is available in **preview** starting from Positron 2025.07.0-204.
 
 Beyond your active files, selected code, and project structure, Positron Assistant is provided with context about your interactive data science work, such as your loaded data, plots, and console history, enabling more relevant guidance for exploratory analysis and modeling.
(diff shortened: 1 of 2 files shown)


assistant.qmd
```diff
@@ -36,6 +36,12 @@ listing:
         link: "https://posit.co/blog/?post_tag=ai-newsletter"
 ---
 
+::: {.callout-note}
+## Looking for Posit Assistant?
+
+Posit Assistant is a new AI coding experience currently available as a preview feature. To learn how to enable it and get set up, visit the [Posit Assistant documentation](https://pos.it/assistant-docs).
```
sharon-wang: Do we want to say a bit more here, and/or link out to the discussion post (once available) with a bit more explanation?

databot.qmd
```diff
@@ -5,14 +5,19 @@ description: "Experimental AI assistant for exploratory data analysis in Python
...
 
 Databot is an AI assistant designed to dramatically accelerate exploratory data analysis for data scientists fluent in Python or R, allowing them to do in minutes what might usually take hours. It’s a complement to the data scientist, not a replacement.
 
+For additional context, see the [Databot introduction blog post](https://posit.co/blog/introducing-databot/) and [Databot is not a flotation device blog post](https://posit.co/blog/databot-is-not-a-flotation-device/).
+
+
+::: {.callout-note}
+Databot is **experimental** and available in research preview; it is not ready for production use.
+:::
+
```
melissa-barca: Did you add this intentionally? It reads a little confusing to call it out as in research preview in the same update we're deprecating it.

sharon-wang: I extracted this from the previous callout and moved it down here as I thought we may still want to note that it's experimental for folks still using it, but maybe it's just too confusing!

Perhaps in the callout where we mention we will deprecate it, I could amend that to say something like "Databot was previously in experimental preview, but will now be deprecated...", and remove this callout?

melissa-barca: Ah, I see! That sounds good to me.

melissa-barca (approved):
LGTM!

PR #345, opened 2026-04-30
Add redirects for migrating our blog to https://opensource.posit.co/blog/

juliasilge:
We, together with the Quarto, tidyverse, Shiny, etc etc etc teams, are moving all our multiple blogs into one centralized spot. This PR sets up redirects now, and then later we can remove the blog section here.


juliasilge:
I don't think the redirects are expected to work on the deploy preview, right? Or do we think I've done something wrong?

sharon-wang:
> I don't think the redirects are expected to work on the deploy preview, right? Or do we think I've done something wrong?

The redirects should work in the preview!

I think they are not working because the blog post files are still getting rendered as part of the site, because we haven't removed the `.qmd` files, so the rendered HTML pages take precedence over the redirects.

I think we should either:
- remove the .qmd files if we are ready
- use a different approach via Quarto (instead of the Netlify redirects), if redirects are supported in the way we need here

juliasilge:
Ah, you are right, @sharon-wang! I guess we'll just go straight to ripping this out as it is the most straightforward way to get the redirects to work.

Just for some visibility, we also have these changes in flight:

- https://github.com/posit-dev/open-source-website/pull/220 which is mostly to stimulate discussion of how we'll get a useful link for us to use
- https://github.com/posit-dev/open-source-website/pull/178 which will add a nice banner to posts being redirected

Hold off on reviewing again until I get some info from the DevRel team.

juliasilge:
OK, I believe we have this set up now on the new website side. 

Check out the deploy preview for:

- https://deploy-preview-345--positron-posit-co.netlify.app/blog/ (this one redirects with a nice banner)
- https://deploy-preview-345--positron-posit-co.netlify.app/blog/posts/2026-04-03-positron-server-jupyterhub/
- https://deploy-preview-345--positron-posit-co.netlify.app/blog/posts/2026-04-06-april-newsletter/

And similar.

Thanks for taking another look now and reviewing, @sharon-wang!

_redirects
```diff
@@ -5,3 +5,11 @@
...
 # /redirect/<redirect-name> <destination-url>
 /redirect/aws-bedrock-iam-permissions  https://docs.posit.co/ide/server-pro/admin/authenticating_users/aws_credentials.html#amazon-bedrock-permissions
+
+# Moving blog over to https://opensource.posit.co/blog/
+/blog  https://opensource.posit.co/blog/q/positron/
+/blog/posts/2026-03-05-outgrow-your-laptop/  https://opensource.posit.co/blog/2026-03-05_outgrow-your-laptop/
+/blog/posts/2026-03-12-notebooks-march-announcement/  https://opensource.posit.co/blog/2026-03-16_notebooks-march-announcement/
+/blog/posts/2026-03-31-python-type-checkers/  https://opensource.posit.co/blog/2026-03-31_python-type-checkers/
+/blog/posts/2026-04-03-positron-server-jupyterhub/  https://opensource.posit.co/blog/2026-04-06_positron-server-jupyterhub/
+/blog/posts/2026-04-06-april-newsletter/  https://opensource.posit.co/blog/2026-04-07_april-newsletter/
```
sharon-wang, line 15: maybe not an issue, but noting that the dates for the notebooks, jupyterhub and april newsletter are a bit different than the original

sharon-wang (approved):
LGTM!

_redirects
sharon-wang: otherwise, I checked all these redirects and they went to the expected place on the new blog ✅

juliasilge: Yes, I'm not entirely sure why the dates ended up different when they got imported over on the new site.

PR #347, opened 2026-04-30
Add docs for the Packages pane

bricestacey:
The Packages pane is a new preview feature for managing R and Python packages in the active session. Add it to the Interpreters section of both Quarto profiles.

I added it to the "interpreters" section...

Screenshot 2026-04-30 at 1 46 46 PM

Screenshot 2026-04-30 at 1 44 16 PM

### DIFF
--- a/packages-pane.qmd
+++ b/packages-pane.qmd
@@ -1,0 +1,52 @@
+---
+title: "Packages Pane"
+description: "Install, update, and uninstall packages in your active R or Python session without leaving Positron. Browse installed packages and search package repositories."
+---
+
+::: {.callout-important}
+The **Packages** pane is a preview feature. It is enabled by default and can be disabled via the [`positron.packages.enable`](positron://settings/positron.packages.enable) setting.
+:::
+
+The **Packages** pane provides an interface for managing the packages installed in your active R or Python session.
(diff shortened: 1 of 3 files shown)


packages-pane.qmd
```diff
@@ -0,0 +1,52 @@
+---
+title: "Packages Pane"
+description: "Install, update, and uninstall packages in your active R or Python session without leaving Positron. Browse installed packages and search package repositories."
+---
+
```
cindyytong, line 5: should we mention somewhere how to set where packages are installed from? 

i.e. setting for positron://settings/positron.r.packageManagerRepository
Image

bricestacey: That is a good idea, will try to start documenting this. We currently defer to the tool used to manage packages.

R: whatever install.packages, pak, or renv use.
Python: whatever pip, uv, or conda use.

This means, any settings to modify this behavior (as shown) should generally "just work" and behave similarly with the packages pane as if they were doing it any other way.

juliasilge:
I would love to get an initial version of this merged today, so it can go in the doc bundle for the 2026.05 release.

Can we log follow-up work for settings that are more about where your packages are installed from and less about the Packages pane itself? Let's move forward with an initial version now.

PR #348, opened 2026-05-01
Update doc for Shiny apps

lionel-:
Follow-up to:
- https://github.com/posit-dev/positron/issues/8450 
- https://github.com/posit-dev/shiny-vscode/pull/108
- https://github.com/posit-dev/positron/pull/12857
- https://github.com/posit-dev/positron/pull/12827

Documents:
- How to configure where to view apps, with the play button or via `runApp()`. The play button configuration also applies to other data apps.
- Brief mention that the play button can be used to debug R Shiny apps.

### DIFF
--- a/develop-data-apps.qmd
+++ b/develop-data-apps.qmd
@@ -9,29 +9,24 @@ Positron provides a simplified method for running interactive data apps via the 
 
 ## Supported app frameworks
 
-Positron supports the following app frameworks:
-
-**Python:**
+Currently, Positron supports the following Python app frameworks:
 
 - [Dash](https://dash.plotly.com/)
 - [FastAPI](https://fastapi.tiangolo.com/)
 - [Flask](https://flask.palletsprojects.com/en/stable/)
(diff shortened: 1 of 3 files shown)


juliasilge (changes requested):
I am going to advocate against trying to support both all the Python data app frameworks and Shiny for R on the same page mixed in together; it's more confusing for _both_ kinds of users this way. I'd like to keep `develop-data-apps.qmd` the way it was before except for adding in the new info on `positron.runApp.previewMode`.

Do you think we should add a new page on Shiny for R? Could be under the "Language Guide" section? Or alternatively we could lean on the RStudio migration settings page as where we surface this for now, until we have more to say? I think you could still link to the info on `positron.runApp.previewMode` on `/develop-data-apps`, or you could make that an [include](https://quarto.org/docs/authoring/includes.html) that is used in two places.

juliasilge (approved):
Thank you so much! I'm really happy with the clarity here 🙌

PR #355, opened 2026-05-06
Use link to new Publisher docs

juliasilge:
The Publisher extension has their docs in a new home here:

https://posit-dev.github.io/publisher/

We can now link to that for folks to get info on that extension.


cindyytong (approved):
ty

PR #362, opened 2026-05-15
Add new skill for updating docs site for regular monthly release

juliasilge:
This skill is intended to be used at this stage of our release process:
https://connect.posit.it/positron-wiki/release-process.html#update-the-downloads-page-and-release-notes-on-the-website

I haven't put it through its paces thoroughly yet, but I'd like to use it next month to see how it goes and if we need to make any changes to it.


juliasilge:
@jonvanausdeln you have been reviewer for most of these PRs. Does this look correct to you?

jonvanausdeln (approved):
LGTM!
Once this is established as working well, we could add another or update this to handle patch releases as well?

juliasilge:
> we could add another or update this to handle patch releases as well?

Yes, totally!

PR #375, opened 2026-05-28
Simplify "Download" button

### LINKED ISSUE #13760: Website: download button should know what platform I'm on
A common feature I observe in other products is that they just give you a download button that works on your platform, you don't have to go hunt for the right version in a list. For example, VS Code: 

Image

But Positron makes you click through to a download page, and then you have to select a build from a list of 10 different platforms and architectures. 

(Also it's odd that we call out "Free Download" on the landing page, makes me think that this is freemium software)

juliasilge:
Fixes https://github.com/posit-dev/positron/issues/13760 in the sense that this addresses the awkward connotations of "Free Download". We all definitely agree that this does not communicate well for our users.

It does _not_ fix the problem of us needing to show both buttons, unfortunately. We can't tell architecture from user agent or anything else in the browser, and we needed to remove our universal installer for a variety of reasons.

### DIFF
--- a/index.qmd
+++ b/index.qmd
@@ -33,7 +33,7 @@ From [Posit PBC](https://posit.co/), the creators of RStudio
 ### Positron unifies exploration and production work in one free, AI-assisted environment, empowering the full spectrum of data science in Python and R.
 
 ::: {.content-visible unless-profile="workbench"}
-<a href="download.qmd" class="btn btn-outline-primary">Free Download</a>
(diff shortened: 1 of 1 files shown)


jthomasmock (approved):
LGTM -- if we wanted to, we could always add clarify in the footer in future or other approaches, but this is a good start!

> Positron™ is `free to use and` licensed under the Elastic License 2.0, a source-available license. Read more about what this license means and our decision to use it.

PR #376, opened 2026-05-28
Change button text from 'Free Download' to 'Download'

nealrichardson:
Oddly, saying "Free" makes you think that Positron is freemium and/or will monetize you in other annoying ways.


nealrichardson:
Ah looks like you're already doing this in #375!

PR #378, opened 2026-05-29
Migrate 4 videos to Git LFS and document LFS setup

sharon-wang:
## Summary
- Re-added 4 videos (`connections.mp4`, `data-explorer.mp4`, `positron-assistant-copytoeditor.mp4`, `positron-hero-fast-tour.mp4`) through the LFS filter. They were committed as raw binary, which caused phantom diffs for anyone with LFS installed.
- Added a "Videos and Git LFS" section to CLAUDE.md and a troubleshooting note to README.md so folks know how to spot and fix this in the future.

## Test plan
- [x] Clone fresh and confirm no phantom diffs on video files
- [x] Verify videos still render correctly on the site


sharon-wang:
Tested fresh clone by running `git clone --branch fix/lfs-video-migration git@github.com:posit-dev/positron-website.git /tmp/positron-website-test` -- `git status`, `git diff` both clean 👍

README.md
```diff
@@ -15,6 +15,8 @@ We currently use the issue tracker for [Positron](https://github.com/posit-dev/p
 
 This repository uses [Git LFS](https://git-lfs.com/) to store large files such as videos. Make sure you have Git LFS installed before cloning.
 
+If you see phantom diffs on video files you didn't change, it usually means those files were committed outside of LFS. Run `git add <file>` to re-add them through the LFS filter, then commit the result.
```
juliasilge: The current wording tells contributors to `git add` and commit the file, but on a fresh clone that actually rewrites the stored blob into an LFS pointer, which is a real change to history rather than local cleanup. A contributor who just wants a clean tree could end up slipping an LFS migration into an unrelated PR without realizing it. Once this PR merges these 4 files are proper pointers, so the phantom diff shouldn't recur for them anyway; the note is most useful as forward-looking prevention.

```suggestion
Because videos are stored in Git LFS, install and initialize Git LFS (`git lfs install`) before adding or replacing a video so it goes through the LFS filter. If you ever see a "phantom diff" on a video file you didn't touch, it means that file was committed as raw binary instead of an LFS pointer. Don't commit the change into an unrelated PR: run `git restore <file>` to clear it locally, and flag the file so it can be re-added through LFS on its own.
```

This keeps the useful breadcrumb, points contributors at "discard and report" instead of "add and commit," and frames it as prevention rather than an expected ongoing state.

CLAUDE.md
```diff
@@ -65,6 +65,10 @@ Key formatting rules:
   - `lint.yml`: URL checking on PRs
   - `publish-release-notes.yml`: Manual workflow for S3/CloudFront deployment
 
+## Videos and Git LFS
+
+Video files (`.mp4`, `.mov`) are stored with Git LFS (see `.gitattributes`). When adding or replacing a video, make sure Git LFS is installed (`git lfs install`) so the file goes through the LFS filter. If a video shows up as changed in `git diff` but you didn't touch it, it was probably committed outside of LFS. Fix it by running `git add <file>` to re-add it through the filter.
+
```
juliasilge, line 80: As written, "Fix it by running `git add <file>` to re-add it through the filter" instructs the agent to stage what is really an LFS migration (it rewrites the stored blob into a pointer, a change to history), and to do it as a side effect of unrelated work. That's behavior we'd want to prevent in an agent.

```suggestion
## Videos and Git LFS

Video files (`.mp4`, `.mov`) are stored with Git LFS (see `.gitattributes`). When adding or replacing a video, make sure Git LFS is installed and initialized (`git lfs install`) so the file goes through the LFS filter.

If a video shows up as changed in `git diff` or `git status` but you didn't touch it, that's a phantom diff: the file was committed as raw binary instead of an LFS pointer, so the clean filter now produces a pointer that differs from the stored blob. Do not stage or commit this as part of unrelated work. Running `git add` would rewrite the stored blob into a pointer, which is a real change to history. Instead, discard it with `git restore <file>`, and handle converting the file to LFS as its own deliberate change.

```

This keeps the diagnostic explanation (which _is_ genuinely useful for the agent!), but steers toward "discard, don't silently commit" and treats the LFS conversion as an intentional, separate change.

juliasilge (changes requested):
Thank you for fixing this! I never saw this problem myself 😱 but I have heard of other people running into it. I dug into why I never saw the phantom diff on my own clone even though I do have Git LFS installed.

The 4 videos really were committed as raw binary, not LFS pointers. On `main`, `git cat-file -p main:videos/connections.mp4` returns the raw MP4 bytes (`ftypisom...`), and `git lfs ls-files` lists the other 15 videos but not these 4. So when the LFS clean filter runs, it produces a small pointer that doesn't match the large raw blob stored in the commit, which is the phantom diff (`Bin 564026 -> 131 bytes`).

Whether you actually *see* that diff comes down to the Git index, not whether LFS is installed:

- Git records `size: 0` in the index for any path that goes through a filter, which forces it to re-run the clean filter on every `git status`. A fresh clone today (with `.gitattributes` present) builds the index that way, runs the filter, and surfaces the phantom diff.
- An older local clone has these entries from before they were properly flagged for LFS, so the index stored the real size (564026) and Git takes the stat-cache shortcut: it never re-runs the filter and reports the file as clean.

I verified this by forcing Git to rebuild those index entries (simulating a fresh clone), at which point the phantom diff appeared on my machine too. So "I have LFS installed" isn't the deciding factor; a stale-but-lucky index in an existing clone is why some of us never noticed it while a fresh clone hits it immediately.

sharon-wang:
@juliasilge I've reworked the guidance so that it's clear that the videos should be fixed in a separate PR from the dev's changes, and included the itemized git steps on how to fix them. I used your wording suggestions as a base and slightly reordered the README since the steps for fixing the video/lfs are hopefully not commonly needed after these remaining files are updated to lfs.

juliasilge (approved):
Thank you!

PR #382, opened 2026-06-03
Remove bullet about "no AI in RStudio"

juliasilge:
Now that RStudio has Posit Assistant in it, we should remove this bullet.


sharon-wang (approved):
LGTM!

PR #383, opened 2026-06-03
Bind checkbox and label in download agreement

jmcphers:
When accepting the download agreement, you need to click very precisely on the checkbox itself, since the `<label>` element is not associated with the checkbox.

This change associates the label with the checkbox, which improves accessibility and also makes it possible to click on the "I agree" text instead of the checkbox itself to accept the terms.


juliasilge (approved):
Thank you! Works great 🙌

PR #384, opened 2026-06-10
update homepage copy

cindyytong:
Update homepage copy to more closely match posit.co site 

Screenshot 2026-06-10 at 9 48 27 AM

workbench profile
Screenshot 2026-06-10 at 9 48 50 AM
Screenshot 2026-06-10 at 9 49 17 AM

PR #388, opened 2026-06-15
Improve how hero headline hyphenated compound words render

juliasilge:
A followup to https://github.com/posit-dev/positron-website/pull/384

The homepage headline ("The data science IDE. Human-driven, AI-assisted.") wrapped awkwardly at _most_ widths. Because a hyphen is a valid line-break point, the browser would split hyphenated compounds mid-word (e.g. "AI-" / "assisted"), and the break landed inconsistently.

This PR:
- Forces the line break at the sentence boundary with `<br>`, so the H1 reads as two clean lines: "The data science IDE." / "Human-driven, AI-assisted."
- Wraps "Human-driven," and "AI-assisted." in `.nowrap` spans (with a new `white-space: nowrap` rule scoped to `.index-section h1`) so those compounds never split mid-word, including when the second sentence wraps on narrow mobile screens.

The H1 sits above the profile-specific content blocks, so this applies to both the public and Workbench builds.

Here's what we'll have after this change for big screens:

Screenshot 2026-06-15 at 8 32 30 AM

For little screens:

Screenshot 2026-06-15 at 8 32 42 AM

### DIFF
--- a/css/theme.scss
+++ b/css/theme.scss
@@ -460,6 +460,11 @@ a.footnote-back {
   font-weight: 600;
 }
 
+// Keep hyphenated compounds in the hero headline from breaking mid-word
+.index-section h1 .nowrap {
+  white-space: nowrap;
+}
+
 .index-section .btn {
   margin: 0.75rem;
   padding: 1rem 2rem;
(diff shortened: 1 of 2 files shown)


cindyytong (approved):
Thank you!!

PR #389, opened 2026-06-17
Assistant: update provider status and setup steps

sharon-wang:
- closes https://github.com/posit-dev/positron/issues/13810
- affects https://github.com/posit-dev/positron/issues/13866 (this PR moves us away from the tabset, but I think we can improve the layout further. However, this PR focuses on updating the content mostly, rather than the layout)

## What Changed

Move the per-provider account, authentication, and settings details out of the getting-started guide into a new assistant-providers.qmd reference page. Getting Started now keeps a single provider table that links into the new page, so the setup flow stays short.

- Add assistant-providers.qmd and register it in both Quarto profiles
- Add Preview/Experimental status badges (CSS in theme.scss), used in the provider table and on the providers page
- Add DeepSeek and Gemini Code Assist as experimental providers
- Repoint completions-page links to the new providers page
- fill in missing/outdated provider configuration instructions, e.g. if settings IDs changed

## What is not yet updated

Will likely land this in a separate PR:

- overrides settings which are getting migrated
- we don't explain how to disable providers now that they're enabled by default
- other settings keys may need to be updated as well


sharon-wang:
Ah, just noticing the CI checks don't run on PRs targeting a branch that isn't main. We'll need to update that first!

juliasilge:
OK, I think I have done this in Netlify:

Screenshot 2026-06-22 at 9 48 14 AM

I added `release/*` as additional targets for "branch deploys", which means we should get deploy previews for PRs to those branches. Shall we push an empty commit here to see if it works?

Here are docs, if you want to check it out: https://docs.netlify.com/deploy/deploy-overview/#branches-and-deploys

juliasilge (changes requested):
Can you run the `/doc-reviewer` skill on the changes in this PR? I notice that a `callout-tip` box has slipped in, which indicates to me that we might need to run that skill to be in compliance with the style guide.

sharon-wang:
Thank you! I've now run the `/doc-reviewer` skill, but had to manually prompt about the callout-tip. I think it's because the skill says "these ABC callouts are allowed" but doesn't explicitly say "callout-tip is NOT allowed".

juliasilge (approved):
This is a huge improvement! We may want to iterate a bit on some of the styling, but I am very much in favor of adding an additional page and making the "getting started" piece feel less overwhelming.

PR #385, opened 2026-06-11
Assistant: update Posit Assistant guidance

sharon-wang:
Closes https://github.com/posit-dev/positron/issues/13934.

Some updates to the docs to provide better guidance around Posit Assistant, noting the deprecation of Positron Assistant, but still including Positron Assistant docs for the time being. This is meant to alleviate some confusion in the meantime, before Positron Assistant is deprecated.

- Added a shared deprecation callout for Positron Assistant pages
- Added "Migrating to Posit Assistant?" tips to Positron Assistant pages to note Posit Assistant related/analogous features 
- Reworked Getting Started to focus more on Posit Assistant, but still give Positron Assistant users the info they need
- Reframed wording across Getting Started and the provider page from "Positron Assistant" to "Positron" / "AI features," and renamed the provider page to "AI Model Provider Privacy Information" (now covers Posit Assistant, Positron Assistant, and Databot).
- Updated Troubleshooting to direct Posit Assistant feedback to the new feedback repo

PR #391, opened 2026-06-19
Update Assistant docs for compliance with the style guide

juliasilge:
I took a look at the changes in https://github.com/posit-dev/positron-website/pull/385 and noticed the use of `.callout-tip`, which the Posit docs style guide forbids. 😅 

I ran the `/doc-reviewer` skill over the files that were changed in that PR and want to propose these current changes. Some of them are not related exactly to the changes in https://github.com/posit-dev/positron-website/pull/385 but I think I'd advocate for making the changes anyway.


sharon-wang (approved):
ah thank you for catching this and making the updates! I checked and I no longer have the `/doc-reviewer` skill after setting up a new machine, so I totally missed this. I'll get that set up!

PR #392, opened 2026-06-23
Prepare release notes for 2026.06.1 patch

juliasilge:
We are doing a patch release for 2026.06, mostly to get an updated version of axios out for folks (the vulnerable version was in Positron itself and the Quarto extension).


jonvanausdeln (approved):
LGTM!

PR #395, opened 2026-06-29
Polish Assistant provider layout

juliasilge:
Closes https://github.com/posit-dev/positron/issues/13866

We'd already removed the tabset, so this PR just does a little polish on this page to bring the badges up into the headers and make the TOC on the right simpler:

Screenshot 2026-06-29 at 1 44 31 PM

### DIFF
--- a/assistant-providers.qmd
+++ b/assistant-providers.qmd
@@ -1,6 +1,7 @@
 ---
 title: "Providers"
 description: "Account requirements, authentication, and settings for each language model provider you can use with Posit Assistant."
+toc-depth: 2
 ---
 
 Positron connects to a range of language model providers. To connect one, run the command _Authentication: Configure Language Model Providers_, select the provider, and authenticate.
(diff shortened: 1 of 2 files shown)


assistant-providers.qmd
```diff
@@ -1,6 +1,7 @@
 ---
 title: "Providers"
 description: "Account requirements, authentication, and settings for each language model provider you can use with Posit Assistant."
+toc-depth: 2
```
juliasilge, line 4: This makes the right-hand TOC more minimal, just a scannable list of the providers.

assistant-providers.qmd
```diff
@@ -145,11 +142,9 @@ To use specific models instead of `model-router`, configure the [`positron.assis
 ]
 ```
 
-Use the model name and deployment identifier from your Foundry resource for each entry. For the full reference on model listings and other model settings, see the [Models](assistant-models.qmd) guide.
+Use the model name and deployment identifier from your Foundry resource for each entry.
```
juliasilge, line 145: Here and below for the custom provider, I remove a link that looks possibly fabricated. Is this meant to link to something in particular?

sharon-wang: ah thanks for catching this! I had played around with a models-specific page at one point, but didn't end up moving forward with it. Thanks for removing this sentence!

sharon-wang (approved):
this looks great!

PR #396, opened 2026-06-30
Add a new GH action to publish release note assets to the CDN

juliasilge:
This PR adds a way to keep release-notes images and GIFs version controlled in this repo and publish them to the CDN automatically, instead of uploading them out of band.

Release notes reference images by absolute CDN URL (`https://cdn.posit.co/positron/releases/release-notes/assets/<name>`), but until now those asset files lived only on the CDN and were never committed here. This PR makes `release-notes/assets/` the source of truth and wires up the upload.

## What's included

- **New workflow `publish-release-notes-assets.yml`**: syncs `release-notes/assets/` to `s3://posit-positron-downloads/positron/<channel>/release-notes/assets/` and invalidates the CloudFront cache for that prefix. It runs automatically on merge to `main` when anything under `release-notes/assets/**` changes, and can also be dispatched manually to pick a channel (`releases`, `staging`, `dailies`; defaults to `releases`). It reuses the same OIDC roles and secrets as the existing publish workflows (`AWS_S3_RW_ROLE`, `AWS_CDN_RW_ROLE`, `AWS_CDN_DISTRIBUTION_ID`).
- **`release-notes/assets/` directory** with a `README.md` documenting the naming convention (`<year>-<month>-<slug>.<ext>`), the CDN URL reference pattern, and how the sync works.
- **`prepare-release-notes` skill**: a guided checklist for authoring `next.md`, including the in-product image preview loop and the steps to move an asset into `release-notes/assets/` before committing. I added this because honestly this process is a big pain in the neck. 😭 
- **CLAUDE.md**: documents the new workflow in the CI/CD section.

## Notes on design

- The sync is additive (`aws s3 sync` with no `--delete`), so it only uploads new or changed files, never removes existing assets, and is safe to re-run.
- This is a public repo, but only collaborators with write access can merge to `main` or trigger `workflow_dispatch`, and fork pull requests can never trigger the workflow or reach the secrets. So no extra approval gate was added.
- A `concurrency` group prevents two quick merges from racing on the same prefix.
- `README.md` is excluded from the sync so it does not land on the CDN.

## Authoring impact

Because assets now publish on merge, you can merge an assets-only change first, get the live CDN URL, and author the release notes against it to preview the real rendering. The in-product local-file preview trick still works for iterating before getting anything up on the CDN.

### DIFF
--- a/.claude/skills/prepare-release-notes/SKILL.md
+++ b/.claude/skills/prepare-release-notes/SKILL.md
@@ -1,0 +1,64 @@
+---
+name: prepare-release-notes
+description: Guides authoring the in-product release notes in release-notes/next.md, especially adding and iterating on images/GIFs with a local in-product preview loop before committing the asset to release-notes/assets/. Use when writing or revising next.md ahead of a release or patch (typically invoked as /prepare-release-notes). The later ship-day promotion of next.md into release.md is handled by /release-update and /patch-update, not this skill.
+---
+
+# Prepare release notes
+
+This skill covers **authoring** `release-notes/next.md`: the content that ships to S3 as the in-product release notes during the release cycle. It is the most painful part of the job when images are involved, because you usually have to iterate on a GIF or screenshot a few times to get it to look right, and previewing an image that is not yet on the CDN takes a specific trick.
+
+By the time `next.md` is promoted into `release.md` on ship day (`/release-update` or `/patch-update`), it should already be final and reviewed. So do the real work here.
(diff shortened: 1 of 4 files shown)


timtmok (approved):
LGTM

We should review this at some point in the future to see if we need to improve the upload times since it'll sync everything in assets. But I don't expect that part of the workflow to increase much with the amount of assets we use in release notes at this time.

PR #406, opened 2026-07-07
Remove some out-of-date limits

jmcphers:
The Quarto inline output docs name some limitations of the feature that are no longer limitations! Remove the stale limitations.


juliasilge (approved):
Thank you!

PR #417, opened 2026-07-17
Add build-and-deploy CI workflow (PR previews + dailies publish)

juliasilge:
This PR adds a new "Build and deploy" GitHub Actions workflow (`.github/workflows/build-and-deploy.yml`), the next piece in moving `positron.posit.co` off Netlify production and onto GitHub Pages, with dev docs served under `/dailies`. This is Step 2 of posit-dev/positron-builds#1127.

The workflow has three jobs:

- **`build`** (on PRs and pushes to `main`): reads the pinned Quarto version, sets up Quarto, renders the `positron` profile, and uploads `_site` as an artifact. On pushes to `main` it also renders the `dailies,positron` overlay and uploads `_site-dailies`. No R or Python is needed in CI because computed pages are served from the committed `_freeze`.
- **`preview`** (PRs only, skipped for fork PRs since they can't access secrets): downloads the `_site` artifact and deploys it to Netlify under a `pr-<number>` alias via `nwtgck/actions-netlify`, which also creates a GitHub Deployment and posts the preview URL as a PR comment.
- **`publish-dailies`** (pushes to `main` only): downloads `_site-dailies` and publishes it to the `gh-pages` branch under `dailies/` with `keep_files: true`, so the released site at the root (and `CNAME`) stays intact.

I need to confirm we have `NETLIFY_SITE_ID` and `NETLIFY_AUTH_TOKEN` repository secrets, pointing at a dedicated Netlify preview site (not production). Until then the `preview` job will fail on PRs.

## Scope and safety

This is still in the prep and early stages of what's outlined in the tracking issue, so we'll land the CI workflow and PR previews while Netlify continues to build and serve production. Nothing user-facing changes at all yet:

- Netlify still builds and serves positron.posit.co.
- The `gh-pages` branch is not served by anything until GitHub Pages is enabled for it (a later phase), so `publish-dailies` just quietly maintains the branch for now.

## Notes

- `keep_files: true` preserves the root site but does not prune: a page removed from the dailies build will linger in `gh-pages/dailies/` on later runs. This is the tradeoff noted in the issue (peaceiris has no "clean only this subdirectory" mode).
- The Quarto version is still read from `netlify.toml` using the same grep already used by `release-docs-bundles.yml`, to keep one source of truth for now. Hoisting that pin into a dedicated file is Step 3.
- After merging, let's plan to watch a few PRs (preview parity) and a few merges (clean dailies publishes) before starting Step 3.

Addresses part of posit-dev/positron-builds#1127.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -1,0 +1,119 @@
+name: Build and deploy
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  # Only cancel superseded runs for PRs; let main-branch publishes finish.
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  build:
+    name: Render docs
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Get Quarto version
+        id: get-quarto-version
+        run: |
+          QUARTO_VERSION=$(grep 'version = ' netlify.toml | sed 's/.*version = "v\?\([^"]*\)".*/\1/')
+          echo "quarto_version=${QUARTO_VERSION}" >> "$GITHUB_OUTPUT"
+
+      - name: Set up Quarto
+        uses: quarto-dev/quarto-actions/setup@v2
+        with:
+          version: ${{ steps.get-quarto-version.outputs.quarto_version }}
+        env:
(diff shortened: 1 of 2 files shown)


juliasilge:
I outlined what we should look for after this merges, to know we are ready to go to step 3: 
https://github.com/posit-dev/positron-builds/issues/1127#issuecomment-5007031741

sharon-wang (approved):
I posted a question re: how to support both dailies docs and docs updates to the release docs: https://github.com/posit-dev/positron-builds/issues/1127#issuecomment-5007373616

Depending on what you think and what approach we want to take, it could affect these changes, or it could be a follow-up

PR #418, opened 2026-07-20
Touch up the python docs

austin3dickey:
This has a bunch of Python-related changes:
- Make the "prerequisites" page a lot shorter since you don't need Python installed before opening Positron now 🎉 
- Link the pages together better
- Call out that venvs outside of the workspace won't be discovered (saw lots of this confusion on intake)
- Moved the ipykernel info from the prerequisites page to the discovery page
- Dropped references to the js locator since we're deprecating that


python-installations.qmd
```diff
@@ -42,16 +48,16 @@ For that reason, we recommend using virtual environments.
 
 ## Automatic uv installation
 
-If you do not have a valid Python available, Positron provides the option to install Python via [uv](https://docs.astral.sh/uv/) to provide an improved Python environment management experience. 
-When Positron starts, it searches for Python on your machine and gives the option install uv if you have no Python or only system Pythons available.
+If you do not have a valid Python available, Positron provides the option to install Python via [uv](https://docs.astral.sh/uv/) to provide an improved Python environment management experience.
+When Positron starts, it searches for Python on your machine and gives the option to install uv if you have no Python or only system Pythons available in the "Start New Console Session" dropdown.
```
juliasilge:
```suggestion
When Positron starts, it searches for Python on your machine and gives the option to install uv if you have no Python or only system Pythons available in the **Start New Console Session** dropdown.
```

I think we should bold this according to the style guide. Can you run the `/doc-reviewer` skill on this file to see how we're doing? Let me know if you've already done this and I am mistaken.

austin3dickey: I constantly forget to run this. 😂

python-installations.qmd
```diff
@@ -27,6 +28,11 @@ Positron searches for Python interpreters in the following locations:
 * `.venv/` and `.conda/` folders in your [workspace](https://code.visualstudio.com/docs/editing/workspaces/workspaces#_singlefolder-workspaces) directory
 * Tool-managed project-specific installations in locations like `~/.local/share/virtualenvs` or where `conda` stores envs
 
+::: callout-important
+Virtual environments outside of the currently-opened folder will not be discovered.
```
isabelizimm: This is true for project-level .venvs, but putting this alone in a callout sort of looks like no interpreters will be discovered outside of the current directory. 

I'd want to be really specific that "Positron will not discover project-level venvs that are declared in a project, unless you have that project open" (but in nicer wording)

python-installations.qmd
```diff
@@ -93,31 +99,29 @@ Positron automatically discovers environments when:
...
+## Other details
 
-Positron only supports Python minor versions 3.9 through 3.14, which are the [currently supported versions](https://devguide.python.org/versions/).
-Other versions may be discovered, but you may run into issues using them.
+### IPython kernel
 
-### Locator type
+Positron requires the [IPython kernel](https://pypi.org/project/ipykernel/) (version 6.19.1 or higher) to communicate with Python.
+By default, the IPython kernel is bundled with Positron for the CPython implementation (the standard Python implementation) and added to the [Python search path](https://docs.python.org/3/library/sys_path_init.html).
+This means you don't have to manually install anything after creating a new environment, unlike working with notebooks in other tools.
```
isabelizimm:
```suggestion
This means you don't have to manually install anything after creating a new environment.
```
a bit confusing to mention notebooks generally in a Python-forward doc

python-installations.qmd
```diff
@@ -93,31 +99,29 @@ Positron automatically discovers environments when:
...
-Positron can use two different discovery mechanisms:
+If you prefer finer control over your Python environment, set the [`python.useBundledIpykernel`](positron://settings/python.useBundledIpykernel) setting to false.
+If you do that or are using an implementation other than CPython, install the IPython kernel yourself with `pip install ipykernel`.
 
-* **Native locator** (default): Uses a native binary for faster discovery
-* **JavaScript locator**: Pure TypeScript implementation
+### Unsupported Python versions
 
-You can control this with the [`python.locator`](positron://settings/python.locator) setting (`"js"` or `"native"`).
+Positron only supports Python minor versions 3.9 through 3.14, which are the [currently supported versions](https://devguide.python.org/versions/).
```
isabelizimm: Hmmm...technically 3.9 is EOL, what do you think about saying something like "Positron supports the currently supported versions. It may continue to support newly EOL versions for a time after" OR just only mention supported versions and not guarantee/mention extra versions

austin3dickey:
Okay I just pushed a commit that
- addressed your feedback
- reorganized the "Discovery locations" section to align with the new environment organization scheme I'll implement as part of https://github.com/posit-dev/positron/issues/8129
- ran the `doc-reviewer` skill on the entire python-installations.qmd page

Can you please take another look? 🙂

python-installations.qmd
```diff
@@ -16,43 +17,66 @@ Positron supports the following Python environment managers:
...
 {{< include "_python-managers.qmd" >}}
 
-Other tools may also be compatible, although they are not officially supported.
+Other tools might also be compatible, although Positron does not officially support them.
 
 ### Discovery locations
 
-Positron searches for Python interpreters in the following locations:
+Positron searches for Python installations in the following locations.
+In general, it is best to use an environment that is highest in this list.
```
isabelizimm, line 25: do we have an article to explain why this is

> it is best to use an environment that is highest in this list

maybe https://realpython.com/python-virtual-environments-a-primer/#why-do-you-need-virtual-environments ?

python-installations.qmd
```diff
@@ -88,43 +112,41 @@ Positron automatically discovers environments when:
...
 
-### Unsupported Python versions
+## Other details
 
-Positron only supports Python minor versions 3.9 through 3.14, which are the [currently supported versions](https://devguide.python.org/versions/).
-Other versions may be discovered, but you may run into issues using them.
+### IPython kernel
 
-### Locator type
+Positron requires the [IPython kernel](https://pypi.org/project/ipykernel/) (version 6.19.1 or higher) to communicate with Python.
```
isabelizimm, line 125: i like this placement way more, thanks!

python-installations.qmd
```diff
@@ -1,9 +1,10 @@
 ---
 title: "Discovering Python Installations"
-description: "Understand how Positron discovers and manages Python installations. Covers automatic uv installation and support for uv, pyenv, conda, venv, pipenv, poetry, and more environment managers."
+description: "Understand how Positron discovers and manages Python installations. Covers automatic uv installation and support for uv, pyenv, conda, venv, and more environment managers."
 ---
 
-This guide explains how Python installations are discovered and managed in Positron.
+This guide explains how Positron discovers and manages Python installations.
+For using Python in Positron, see our [guide to development](guide-python.qmd).
```
juliasilge:
```suggestion
For using Python in Positron, see our [Python language guide](guide-python.qmd).
```

Or something else maybe?

juliasilge (dismissed):
Looks great to me!

python-installations.qmd
austin3dickey: I was actually thinking of making a blog post about this once my reorganization changes are in! Your link is a good resource in the meantime, so I can put it in.

PR #422, opened 2026-07-22
Clarify Positron's prerequisites

### LINKED ISSUE #15036: positron website: it seems as though Python and R are mandatory
On the Positron website, under [Prerequisites](https://positron.posit.co/install.html#prerequisites) for installing Positron, there are mentions of Windows, Python setup, and R setup. There's no mention of the fact that Python/R are optional, and you can use Positron without these languages. It could be a bit confusing for folks who land on that page with no other context for setting up Positron.

(I'll wait to fix this until after https://github.com/posit-dev/positron-website/pull/418 goes in, since that partially addresses it for Python)

isabelizimm:
Related to https://github.com/posit-dev/positron/issues/14848
Fixes https://github.com/posit-dev/positron/issues/15036

Clarify what is needed before installing Positron


install.qmd
```diff
@@ -9,24 +9,34 @@ aliases:
...
+
+Positron runs without any programming language installed. If you plan to use R or Python, these steps get your environment working.
+
+::: {.panel-tabset}
+### Python
 Positron works with [actively supported versions](https://devguide.python.org/versions/#versions) of Python.
 
 If you do not have a valid Python version available, Positron [can install one for you](python-installations.qmd#automatic-uv-installation).
-If you prefer to use an existing environment, Positron makes it easily available to use, provided it meets the [qualifications](python-installations.qmd).
+If you prefer to use an existing environment, Positron supports the following Python environment managers:
```
isabelizimm, line 31: I know this is repeating what is on the Python page, but I feel like having folks go from Downloads page -> Install page -> Python Installations page to see if their environment manager works is too many jumps.

isabelizimm:
~I'm not sure why this is failing...I don't believe these changes would have removed R from our build pipeline?~ Ope, need to update the freeze

install.qmd
dhruvisompura: I think it makes a lot of sense to have everything on one page. I would bet most users aren't doing all of this link navigating (I definitely don't whenever I have to download an app!).

install.qmd
```diff
@@ -9,24 +9,34 @@ aliases:
...
+In most cases, Positron is ready to go as soon as it is installed. Check the section for your platform or language to see if you need anything more.
+
+### Windows setup
 
 If you're using Windows, make sure you have the [latest Visual C++ Redistributable](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#latest-microsoft-visual-c-redistributable-version) installed.
 
-### Python setup
+### Optional language setup
+
+Positron runs without any programming language installed. If you plan to use R or Python, these steps get your environment working.
```
dhruvisompura: minor nit: the wording here is throwing me off. I feel like "these steps get your environment working" should say:
```suggestion
Positron runs without any programming language installed. If you plan to use R or Python, follow these steps to get your environment working.
```

install.qmd
dhruvisompura, line 97: I didn't realize that we also had this nice download section on this page! I feel like this should be what we actually show on our downloads page with all of this install content below it!

dhruvisompura (dismissed):
Overall the changes look good to me. Had a minor nit but not blocking.

install.qmd
juliasilge: FWIW we _do_ link to the `/install` page _from_ the downloads page. I have often wondered if we've got this organized well, and if we need to have the release notes at all on the downloads page.

juliasilge (dismissed):
Thank you! I like the suggestions that @dhruvisompura left. 👍

install.qmd
isabelizimm: oh thank you! i was fighting the `doc-reviewer` on this line and this was our compromise 😆 but I like your suggestion way better!

install.qmd
isabelizimm: It does feel a bit disjointed to have the two pages. I don't think I would have a strong intuition of whether I should click on "download positron" or "install positron" from this list. My guess is that folks simply download from whatever page they land on first 😅 I'll merge this in, but it might be worth thinking about rearranging?

Screenshot 2026-07-23 at 9 40 54 AM

dhruvisompura: I definitely think its worth rearranging! This could be something the onboarding workstream takes on.

An ideal state to me would be something along the lines of:
combining the Install+Download into a single page, and then moving the Downloads and Release Notes as links in the top level nav bar. That's more on par with how other sites do it and its a lot less confusing. Its a much nicer UX when folks can find everything they need to get started on one page.

I would also vote to show the simplified downloads button for your OS at the top of the downloads page and then have the list of all builds under a collapsable "Show More" table.

I'll see how we feel about this idea and file an issue accordingly.

PR #425, opened 2026-07-28
feat: publish slim LLM-only docs bundles

midleman:
### Summary
Adds a `build-llms-bundle.sh` script that packages LLM-friendly `.llms.md` docs into a versioned zip bundle into our existing release-docs workflow.

### The flow
1. Render both Quarto profiles (unchanged)
2. **Build the LLM bundles**, once per profile
   - 2a. Copy `llms.txt` and the `*.llms.md` docs into a staging dir
   - 2b. Rewrite the site URLs in `llms.txt` to relative paths
   - 2c. Guards: no site URL left, all links relative, all targets exist, file count matches
   - 2d. Write `bundle.json`
   - 2e. Zip, then check the zip's contents against what was staged
   - 2f. Check the zip against the 1MB size budget
   - 2g. Write and verify the `.sha256sum`
3. Configure AWS credentials
4. Pre-flight: stop if this version is already published, unless `overwrite`
5. Build the full docs zips (unchanged)
6. Cut the GitHub release, marked prerelease unless the channel is `releases`
7. Upload the full docs zips (unchanged)
8. Upload the LLM zips and checksums, immutable, checksum first
9. Move the `latest` aliases, `releases` channel only
10. Assert: every object exists, right Cache-Control, aliases hold this release's bytes
11. Invalidate CloudFront (unchanged)

### QA Notes
I did as much local testing as I could - see [this PR comment](https://github.com/posit-dev/positron-website/pull/425#issuecomment-5122408082) below.

#### Actions to perform post-merge

<details>
<summary>Post-merge checklist</summary>

- [ ] **Staging run** to exercise the pipeline for real:
  - Run `workflow_dispatch` with:
    - `release_channel: staging`
    - `version: 0.0.0-llmstest`
    - `create_release: true`
  - Aliases are gated on `releases`, so staging covers build, upload, and assert without touching anything live.
  - (`create_release: false` routes to `gh release upload`, which fails on a tag that does not exist.)

- [ ] Delete the throwaway release and tag:

  ```bash
  gh release delete 0.0.0-llmstest --yes --cleanup-tag
  ```

- [ ] Pull the staging bundle back down and spot-check:
  - `shasum -a 256 -c` against its sidecar
  - `bundle.json` fields
  - Confirm the workbench bundle excludes:
    - `download`
    - `install`
    - `privacy`

- [ ] **Verify `Cache-Control` and conditional GET through the CDN** (commands below).
  - Cache headers are only verified at the S3 origin.
  - Moto cannot test CloudFront.
  - This is the one behavior not covered by automated tests.

- [ ] **First `releases` run**
  - Confirm the four alias objects exist.
  - Verify each alias digest matches its versioned ZIP.
  - The assert step now performs this verification, so a green run confirms success.

- [ ] **Consumer check**
  - `parseManifest()` accepts the bundle on a cold-cache Positron install.
  - Assistant reads documentation from disk without fetching from `positron.posit.co`.

- [ ] Record the rollback procedure.
  - Versioned objects are immutable.
  - Recovery consists of:
    - Repointing aliases to the previous version with `aws s3 cp`
    - Using `--cache-control no-cache`
    - Performing a CloudFront invalidation
  - Never overwrite a published versioned object.

### CDN checks

```bash
BASE=https://cdn.posit.co/positron

# Does CloudFront forward the Cache-Control we set at the origin?
curl -sI $BASE/staging/docs/positron-llms-0.0.0-llmstest.zip \
  | grep -i 'cache-control\|etag\|x-cache'

# no-cache is only worth setting if this returns 304
ETAG=$(curl -sI $BASE/releases/docs/positron-llms-latest.zip \
  | grep -i etag \
  | tr -d '\r' \
  | cut -d' ' -f2)

curl -sI -H "If-None-Match: $ETAG" \
  $BASE/releases/docs/positron-llms-latest.zip \
  | head -1
```

</details>

### DIFF
--- a/scripts/build-llms-bundle.sh
+++ b/scripts/build-llms-bundle.sh
@@ -46,7 +46,10 @@ cp "$SITE_DIR/llms.txt" "$STAGE/llms.txt"
 # and exits non-zero while BSD tar exits 0, so without this the same broken
 # render fails cryptically on CI and silently on a Mac. A bundle of llms.txt
 # plus bundle.json and no docs is useless either way.
-DOC_COUNT="$(cd "$SITE_DIR" && find . -name '*.llms.md' -type f | wc -l | tr -d ' ')"
+# Counting NUL bytes from -print0 rather than lines from `wc -l`: a filename
+# containing a newline would otherwise inflate the count and trip the staged-count
+# comparison at step 5 with a mismatch that points at the wrong cause.
+DOC_COUNT="$(cd "$SITE_DIR" && find . -name '*.llms.md' -type f -print0 | tr -dc '\0' | wc -c | tr -d ' ')"
 if [ "$DOC_COUNT" -eq 0 ]; then
 	echo "error: no *.llms.md files under $SITE_DIR; the Quarto render produced no LLM docs." >&2
 	exit 1
@@ -58,21 +61,28 @@ fi
 # 2. Rewrite llms.txt to bundle-relative paths. This is what schema 1 promises.
 # Write-and-move rather than `sed -i`: in-place editing needs no suffix on GNU
 # sed and a mandatory one on BSD, so no single `-i` spelling is portable.
-sed "s|https://positron\\.posit\\.co/||g" "$STAGE/llms.txt" > "$STAGE/llms.txt.rewritten"
+# `-E` with an optional trailing slash so the bare origin is stripped too:
+# site-url carries no trailing slash, so both forms appear in practice.
+sed -E "s|https://positron\\.posit\\.co/?||g" "$STAGE/llms.txt" > "$STAGE/llms.txt.rewritten"
 mv "$STAGE/llms.txt.rewritten" "$STAGE/llms.txt"
 
-# 3. Guard: no bundled file may still reference the site. Paths are reported
-# relative to $SITE_DIR, since a mktemp path is not actionable for an operator.
-if grep -rl 'positron\.posit\.co' "$STAGE" | sed "s|^$STAGE/|$SITE_DIR/|" | grep . ; then
-	echo "error: bundled files still reference positron.posit.co (listed above)." >&2
-	echo "The rewrite assumes only llms.txt carries site links. That assumption broke." >&2
+# 3. Guard: llms.txt must no longer reference the site. Scoped to llms.txt
+# because that is the only file step 2 rewrites. The .llms.md docs legitimately
+# carry absolute site links in prose -- release notes link to doc pages -- and
+# those stay valid URLs for a reader who has a browser, so failing on them would
+# turn ordinary docs authoring into a broken release.
+if grep -n 'positron\.posit\.co' "$STAGE/llms.txt" ; then
+	echo "error: llms.txt still references positron.posit.co (listed above)." >&2
+	echo "The rewrite in step 2 did not cover every form of the site URL." >&2
 	exit 1
 fi
 
-# 4. Guard: every link in llms.txt must now be bundle-relative. Matches any
-# scheme case-insensitively, scheme-relative `//host`, and root-relative `/path`
-# - none of which resolve inside an extracted bundle.
-if grep -oE '\]\([^)]+\)' "$STAGE/llms.txt" | grep -Ei '\(([a-z][a-z0-9+.-]*:)?//|\(/' ; then
+# 4. Guard: every link in llms.txt must now be bundle-relative. Rejects any
+# scheme case-insensitively (`https:`, and also non-slash ones like `mailto:`,
+# which would otherwise reach step 5 and be reported as a missing file),
+# scheme-relative `//host`, and root-relative `/path` - none of which resolve
+# inside an extracted bundle.
+if grep -oE '\]\([^)]+\)' "$STAGE/llms.txt" | grep -Ei '\((([a-z][a-z0-9+.-]*:)|//|/)' ; then
 	echo "error: llms.txt contains links that are not bundle-relative (listed above)." >&2
 	exit 1
 fi
@@ -85,7 +95,10 @@ fi
 # self-consistent whatever is missing.
 MISSING_LINKS=0
 while IFS= read -r target; do
+	# Fragment first, then query: a link is path?query#fragment, and neither
+	# suffix is part of the filename on disk.
 	target="${target%%#*}"
+	target="${target%%\?*}"
 	[ -n "$target" ] || continue
 	if [ ! -f "$STAGE/$target" ]; then
 		echo "error: llms.txt links '$target', which is not in the bundle." >&2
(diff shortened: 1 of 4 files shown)


midleman:
## What I was able to test

### 1. Fixture tests (in CI, on every PR)

`scripts/test-build-llms-bundle.sh` covers 23 cases: both profiles, the artifact shape (zip entries, `fileCount` vs. the real entry count, schema fields, digest verification), and every guard's rejection path. It runs in the new `scripts` job in `lint.yml` next to `shellcheck`.

### 2. Real Quarto render (local, 1.9.23)

Rendered both profiles with the version CI pins, then built bundles from the actual output:

| profile | docs | bundle | digest |
| --- | --- | --- | --- |
| positron | 96 | 98 files, 258 KB | verified |
| workbench | 82 | 84 files, 207 KB | verified |

The two profiles emit different `llms.txt` link forms, because workbench has no `site-url`:

```
positron    - [Welcome](https://positron.posit.co/welcome.llms.md)
workbench   - [Welcome](welcome.llms.md)
```

Both work. On workbench the rewrite is a no-op and the links are already bundle-relative.

This also caught a real bug. The site-link guard originally scanned every staged file, but only `llms.txt` gets rewritten, so any doc carrying an absolute site link in prose failed the build: 13 files on positron, 11 on workbench, including `databot`, `privacy`, and every dated release-notes page. That would have failed the first release run. The guard is now scoped to `llms.txt`, and prose links in docs stay valid URLs.

### 3. Full S3 flow against moto (local)

Ran the S3 steps against `moto_server`, with the real `aws` CLI pointed at it via `AWS_ENDPOINT_URL`. The harness pulls the `run:` bodies straight out of the workflow YAML and executes them in order, so it can't drift from what the workflow actually does.

Happy path: all five steps green, 8 objects published, and `Cache-Control` round-trips as intended (`immutable` on versioned objects, `no-cache` on aliases).

Then I broke things on purpose, since a green run says nothing about whether the assertions have teeth:

| injected fault | result |
| --- | --- |
| re-run same version, no `overwrite` | pre-flight blocks it |
| same, with `overwrite=true` | proceeds |
| alias given `immutable` instead of `no-cache` | cache-control check fails it |
| alias repointed at the wrong bundle | digest check fails it |
| sidecar deleted | existence check fails it |

Also ran the `staging` path: uploads and assertions pass, and zero `-latest` aliases get created.

### Not tested

- **CloudFront.** The `Cache-Control` values are verified at the origin only. Whether CloudFront forwards them, and whether `no-cache` really produces the conditional-GET 304 that the design depends on, needs the real staging run.
- **moto is a reimplementation, not S3.** A pass means the commands are well-formed and the logic holds, not that real S3 behaves identically.
- **The GitHub release steps and the two OIDC credential steps** (both pre-existing and unchanged here).
- **Step-level `if:` conditions.** The harness runs `run:` bodies, so I applied the alias step's `release_channel == 'releases'` gate by hand. The in-script channel gate in the assert step is genuinely covered.


---

## Built artifacts

Both built locally from the Quarto 1.9.23 render, version `2026.07.1-5`:

| | positron | workbench |
| --- | --- | --- |
| files | 98 | 84 |
| size | 258,363 B | 207,456 B |
| layout | flat + `release-notes/` (12) | flat + `release-notes/` (12) |

```
sha256  2eb558ff69923bf5a6c2c61924698450ea57684f74bfb5bd1fe5907acedafbc6  positron-llms-2026.07.1-5.zip
sha256  ea9f59f06123f852f7a3dd2c75f2f44b52eda993ae19e729ab82570259b9e76a  positron-workbench-llms-2026.07.1-5.zip
```

<details>
<summary><code>bundle.json</code> from each bundle</summary>

```json
{
  "schema": 1,
  "profile": "positron",
  "version": "2026.07.1-5",
  "generated": "2026-07-29T19:03:00Z",
  "docsBaseUrl": "https://positron.posit.co/",
  "fileCount": 98
}
```

```json
{
  "schema": 1,
  "profile": "workbench",
  "version": "2026.07.1-5",
  "generated": "2026-07-29T19:03:00Z",
  "docsBaseUrl": "https://positron.posit.co/",
  "fileCount": 84
}
```

</details>

<details>
<summary>Profile exclusions, verified by diffing the two bundles</summary>

Present in positron, absent from workbench:

```
add-to-path.llms.md
dev-containers.llms.md
download.llms.md
education.llms.md
install.llms.md
licensing.llms.md
privacy.llms.md
remote-ssh.llms.md
tutorial-get-started-ipynb.llms.md
tutorial-get-started-quarto.llms.md
tutorial-migrate-from-RStudio.llms.md
tutorial-migrate-from-vscode.llms.md
tutorials.llms.md
updating.llms.md
```

That is 14 pages, matching the `render:` exclusion list in `_quarto-workbench.yml` exactly. Nothing appears in workbench that is missing from positron.

</details>

<details>
<summary>Link rewriting, before and after</summary>

Rendered `_site/llms.txt` (positron has <code>site-url</code> set):

```
- [Welcome](https://positron.posit.co/welcome.llms.md)
- [Updating Positron](https://positron.posit.co/updating.llms.md)
```

Same file inside the bundle:

```
- [Welcome](welcome.llms.md)
- [Updating Positron](updating.llms.md)
```

95 links, all resolving to real files in the bundle, and no `positron.posit.co` left in the index. Doc bodies keep their absolute prose links on purpose: 13 `.llms.md` files still contain them, which is what `docsBaseUrl` in the manifest is for.

</details>

juliasilge (approved):
This looks correct to me! 🙌

scripts/build-llms-bundle.sh
```diff
@@ -0,0 +1,228 @@
...
+cat > "$STAGE/bundle.json" <<JSON
+{
+  "schema": ${SCHEMA},
+  "profile": "${PROFILE}",
+  "version": "${VERSION}",
+  "generated": "$(date -u +%Y-%m-%dT%H:%M:%SZ)",
+  "docsBaseUrl": "${DOCS_BASE_URL}",
+  "fileCount": $((FILE_COUNT + 1))
+}
+JSON
```
dhruvisompura, line 172: What is the `bundle.json` file used for?

midleman: it's the one file in there that describes the bundle rather than being docs. nothing reads it _yet_ - Positron will, in the client-side half of this work.

- `schema` is the compatibility gate. A reader checks it first and can bail cleanly rather than misinterpreting contents it doesn't understand. 
- `profile` and `version` say which docs set this is and which release it describes, so the client can tell whether what it has on disk is what it wants.
- `fileCount` is a self-check after extraction: count the files, compare. Catches a partial or truncated extract.
- `docsBaseUrl` is the prefix step 2 strips out. Keeping it means a consumer can turn a bundle-relative path back into a live URL, so a local doc can still link out to the web page.

PR #426, opened 2026-07-30
Clarify out-of-the-box support (R, Python) plus extensions

juliasilge:
This PR updates the page at <https://positron.posit.co/managing-interpreters.html> to clarify that Positron:

- comes with full support for Python and R out of the box
- can be extended to support other language interpreters with extensions

I used two examples here, the Julia extensions plus ggsql.

Related to https://github.com/posit-dev/positron/issues/3679

### DIFF
--- a/managing-interpreters.qmd
+++ b/managing-interpreters.qmd
@@ -3,7 +3,11 @@ title: "Managing Interpreters"
 description: "Manage multiple Python and R interpreter sessions in Positron. Switch between environments, configure startup options, and run multiple sessions simultaneously."
 ---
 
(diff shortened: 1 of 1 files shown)


isabelizimm (commented):
The prose LGTM, but I don't think callouts are included in search, eg, if I look up Julia I only see the reference in the extensions development page. We may want to pull this onto the page itself, so it's more discoverable.

Image

juliasilge:
I totally forgot that about the callout boxes; thank you very much, @isabelizimm!

Screenshot 2026-07-30 at 10 55 14 AM

isabelizimm (approved):
🚀

PR #428, opened 2026-07-31
Publish site videos to cdn.posit.co

juliasilge:
Related to https://github.com/posit-dev/positron-builds/issues/1127

I was getting ready for us to take the next step to move to GitHub Pages, and I realized that the way we host videos is a BLOCKER right now, to that, because it's so much bandwidth that it would put us way over the GH Pages limit. Maybe it was never a good idea to host the videos the way we did, but it seemed like the best option at the time. We've got some nice infrastructure in this repo now to get things up on to the CDN (originally set up for the release notes assets) so we can do something similar here.

I will follow up this PR with changes to the site, after the videos themselves are available from their new homes. The "poster" images are because I'm also changing these videos to not auto-load, to help with page load and also bandwidth.


jonvanausdeln (approved):
LGTM!

PR #429, opened 2026-07-31
Use our new CDN hosting for the videos

juliasilge:
Related to:

- https://github.com/posit-dev/positron-builds/issues/1127
- https://github.com/posit-dev/positron-website/pull/428

Our videos now come from `cdn.posit.co` instead of Netlify, so we can move forward with a transition to GitHub Pages within the bandwidth restrictions. We already published all 18 videos and their poster frames in #428, so the URLs here resolve today and the preview should play.

What this PR does in addition to that:

- A custom `cdn-video` shortcode replaces Quarto's built-in `{{< video >}}` at 18 call sites. The built-in one injects video.js into `<head>`, 569 KB of render-blocking JavaScript with no opt-out, for a player we use none of.
- One line in `_quarto.yml` keeps the video files out of every profile's output. Quarto concatenates `project.resources` across profiles, so a single top-level entry covers positron, workbench, and dailies. This also stops the positron docs bundle from shipping 18 broken LFS pointer stubs.
- The build no longer checks out LFS, since nothing reads the video bytes now. `publish-videos.yml` keeps its LFS checkout.

Both the video and poster URLs are written out in full at each call site, on purpose, so what you read in the source is what the browser requests. The Workbench "no videos" rule moved out of a separate Lua filter (`_extensions/video-filter/`, deleted) and into the shortcode, which returns `pandoc.Null()` before building any HTML.

When you review, please render both profiles, and probably also start clean, since stale output gets copied into `_site` and skews the greps:

```bash
rm -rf _site _site-workbench
QUARTO_PROFILE=positron quarto render
QUARTO_PROFILE=workbench quarto render
```

In `_site`, the files should be gone but the videos should still play. The poster frame should appear immediately, with no black rectangle first.

In `_site-workbench`, grepping for `<video`, `videojs`, `.mp4`, and `poster=` should all return 0 matches, with no `videos/` directory.


juliasilge:
In c2f601afcd3ac88f41be558cd58d495d882c547c I cut the homepage hero from 22.5 MB to 10.4 MB, a 52% reduction, by going from 1920x1080/60fps to 1600x900/30fps and dropping the audio track. The audio was digitally silent (mean and max volume both at -91 dB), so it was 952 KB of nothing. The new video looks fine to me on my big monitor.

The filename is stable, so no `.qmd` changes are needed and `publish-videos.yml` picks it up on merge. The poster is regenerated from the new source and lands on the same frame. Note the one week max-age, which means browsers holding the old file can serve it until it expires, so the bandwidth win phases in rather than landing at once.

videos/README.md
```diff
@@ -0,0 +1,98 @@
+# Videos
+
+Videos embedded in the site live here so they are version-controlled alongside
+the pages that use them. They are stored with Git LFS (see `.gitattributes`), so
+make sure LFS is installed and initialized (`git lfs install`) before adding or
+replacing one.
+
+The site does **not** serve these files. Netlify excludes them from the build
+output (`"!videos/**"` in `_quarto-positron.yml` and `_quarto-workbench.yml`),
```
dhruvisompura: Looks like this needs to be updated to say:
```suggestion
output (`"!videos/**"` in `_quarto.yml`),
```

tutorial-migrate-from-RStudio.qmd
```diff
@@ -143,7 +143,9 @@ Open the Command Palette and search for _Preferences: Color Theme_. Use the arro
 
 The video below shows how to preview color themes with _Preferences: Color Theme_.
 
-{{< video videos/tutorial-migrate-color-themes.mp4 aria-label="Use _Preferences: Color Theme_ to preview color themes." >}}
+{{< cdn-video https://cdn.posit.co/positron/releases/videos/tutorial-migrate-color-themes.mp4
+  poster="https://cdn.posit.co/positron/releases/videos/tutorial-migrate-color-themes-poster.jpg"
+  aria-label="Use _Preferences: Color Theme_ to preview color themes." >}}
```
dhruvisompura, line 148: really minor nit: 
I don't think its important enough to fix this but just wanted to flag that some videos don't look embedded in the page because the recording doesn't match the positron window size so there's a black background:

Image

 and video in `positron-notebook-editor.qmd` for example

Image

I wonder if we should add a note about video settings/preferences in the README.md and CLAUDE.md file? I haven't recorded any videos for our site so I don't have a suggestion on how to do that but maybe Claude knows 😅

dhruvisompura (dismissed):
This looks great to me! The `<video>` element created in the lua script for the shortcode had all the right attributes, etc.

There is one small change needed in the README but otherwise ready to merge. This change doesn't need a re-review to get merged in. The other comment was more an observation and not something we need to fix but something to keep in mind if we take future videos.

PR #433, opened 2026-08-03
Publish released docs to the `gh-pages` root

juliasilge:
Implements step 3 of posit-dev/positron-builds#1127 where we publish the released docs to the root of `gh-pages`, and also consolidates the Quarto version pin.

## What this PR does

- **`publish-site` job in `release-docs-bundles.yml`.** Mirrors the rendered `_site` onto the root of `gh-pages` using `JamesIves/github-pages-deploy-action` with `clean: true`, so pages deleted from the docs stop being served. `clean-exclude` carves out the two root entries this job does not own: `dailies/`, republished on every merge to `main` by `build-and-deploy.yml`, and `CNAME`, which GitHub writes when the custom domain is set. The job is gated on `release_channel == 'releases'`, so dailies and staging runs keep building S3 bundles without touching the site people read.

- **Shared `gh-pages-deploy` concurrency group.** There are now two workflows writing to `gh-pages` and both push with `--force`, so a release publish overlapping a merge to `main` could otherwise drop the other's commit. The group queues them rather than cancelling, so neither publish is lost.

- **Optional `ref` input, defaulting to `main`.** The checkout previously hardcoded `main`, which meant dispatching the workflow against another branch silently published `main` anyway. This makes the backport-branch procedure for fixing the already-released site actually work.

- **`.quarto-version` as the single pin.** Both workflows read it instead of grepping `netlify.toml`. Netlify cannot read a pin out of another file, so `netlify.toml` keeps its own copy while it still builds production, and the `build` job now fails if the two disagree. Both the guard and the `netlify.toml` pin go away at cutover from Netlify to GH Pages.

There is still a bit of work to do to get us all the way to GH Pages, but I think this should be all we need to do before the release Thursday, to confirm things are working.

## State before the release

GitHub Pages is now enabled on this repo, serving `gh-pages` at `/`, HTTPS enforced, no custom domain. The [first build succeeded in 40s](https://github.com/posit-dev/positron-website/deployments/github-pages).

- <https://posit-dev.github.io/positron-website/> returns 404. This is expected because the `gh-pages` root holds only `.nojekyll` and `dailies/`, with no `index.html` until the first release publish.
- <https://posit-dev.github.io/positron-website/dailies/> returns 200 with assets and `search.json` resolving.

Quarto emits relative asset paths throughout (it even rewrites the root-absolute footer logo to `./images/...`), so the project-page prefix is not a problem and the github.io URL is a faithful preview, not just a structural check.

Netlify is untouched and still serves positron.posit.co, so nothing user-facing changes when this merges.

## What to check after Thursday's release

The `publish-site` job has never run. Dispatch **Release Documentation Bundles** with `release_channel: releases` as usual, then:

- [ ] `https://posit-dev.github.io/positron-website/` serves the released docs and renders correctly (nav, search, images, CSS).
- [ ] `https://posit-dev.github.io/positron-website/dailies/` still returns 200. **This is the important one.** It confirms `clean-exclude: dailies/` protected the dev docs from the root clean, and it is the one behavior verified only by reading the action's `rsync` invocation rather than by running it.
- [ ] The footer at the root shows the release version, and the footer under `/dailies/` still shows the dev version with the announcement banner.
- [ ] `git ls-tree --name-only origin/gh-pages` shows the site at root plus `dailies/` and `.nojekyll`.
- [ ] The `publish-site` job actually ran (it is gated, so confirm it was not skipped) and the `release-site` artifact was produced by `build-docs`.
- [ ] positron.posit.co is unchanged and still served by Netlify.

Then, on the next merge to `main` after the release:

- [ ] `publish-dailies` still succeeds and does not disturb the newly published root.
- [ ] The dailies run does not upload a `release-site` artifact, confirming the `releases` gate on the upload.

## If something goes wrong

Blast radius is zero, because DNS still points at Netlify and positron.posit.co is not served from `gh-pages` yet. A bad root publish is only visible at the github.io URL. If `dailies/` were somehow clobbered, the next merge to `main` republishes it.


jonvanausdeln (approved):
LGTM!

PR #435, opened 2026-08-03
Consolidate the install and download pages, and surface release notes in the nav

dhruvisompura:
This PR merges the prerequisites from `install.qmd` into `download.qmd` and deletes the install page, so getting Positron running lives on one page instead of two. 

I also moved the Release Notes page out of About where it was pretty hidden and made it a too navbar entry. It's a full width page now like `features.qmd` and `tutorials.qmd`.

Some redirect handling: `/install` and `/start` get their own redirect rules alongside the `.html` forms. Netlify matches redirect sources literally, so a rule for `/install.html` does not cover `/install`, and both return 200 today.

Note: the `## Positron Pro on Posit Workbench` section still sits at the bottom of the download page. I am considering moving it under About, next to the education and licensing content but wasn't sure if it was on this page for a reason. Would it be better off under About, or is the download page the right place for this?

### Validation Steps

1. Visit `/install`, `/install.html`, `/start`, and `/start.html`. Expected: each lands on the download page.
2. Visit `/install.html#prerequisites`. Expected: lands on the download page, scrolled to **Prerequisites**.
3. Load the download page. Expected: the hero detects your OS, the matching card shows a **Your OS** badge, and the console is clean.
4. Expand **SHA-256 checksums**. Expected: 10 rows, and clicking a hash copies the full 64-character value.
5. Check the navbar. Expected: **Release Notes** sits right of Guides, and that page renders full width with no sidebar.
6. Visit `/release-notes.html#latest-release`. Expected: still resolves.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


juliasilge:
I made some changes here, @dhruvisompura, for you to take a look at:

- I handled the redirects _within_ the Quarto site using `aliases`, because of what we're doing with #436 
- I changed the sidebar nav for the release notes, so it is more consistent with the rest of the site
- I added the new modal we need for affirmative consent, for the license acceptance

I confirmed that we are OK with the Plausible metrics with this new approach.

juliasilge (approved):

dhruvisompura:
> I made some changes here, @dhruvisompura, for you to take a look at:
> 
> * I handled the redirects _within_ the Quarto site using `aliases`, because of what we're doing with [Generate static redirect stubs for GitHub Pages #436](https://github.com/posit-dev/positron-website/pull/436)
> * I changed the sidebar nav for the release notes, so it is more consistent with the rest of the site
> * I added the new modal we need for affirmative consent, for the license acceptance
> 
> I confirmed that we are OK with the Plausible metrics with this new approach.

Thank you for making these changes! I just reviewed the changes in the new preview and they look good to me.

PR #438, opened 2026-08-05
Reframe daily builds note around early feedback

testlabauto:
## Summary

Reworks the second bullet in the "dailies" section of `updating.qmd`. The previous wording led with the QA caveat ("only go through automated testing... not our full QA process") and framed dailies as being for "adventurous early adopters." The new wording leads with the benefit — trying capabilities as they're developed — and notes that user feedback shapes those features before they reach everyone.

The neighboring bullets still cover the practical expectations (near-daily updates, and what to do if a build is broken), so the caveat isn't the only signal readers get.

## Test plan

- [ ] Netlify preview renders `/updating` correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nealrichardson (approved):
:shipit:

juliasilge (approved):
Thank you!

PR #445, opened 2026-08-06
Fix up the social sharing metadata

juliasilge:
The home page has been serving a broken Open Graph image. Since #170 `index.qmd` has been trying to write its own `og:image` and other tags in an `include-in-header` block, pointing at `/images/index-positron-social-share.png`, but that never worked. This has been broken for about a year!

That image was a poor choice anyway. Despite the name, it is a diagram of the user interface, with callout boxes that label the Activity bar, Primary side bar, Editor, and Panel. Slack and X show previews at about 360px wide, where the screenshot text is unreadable. This PR removes the hand-written block and lets `open-graph: true` do the work, turns on `twitter-card` (the site had no Twitter Card tags at all), and removes the unused file. The home page and interior pages now send one set of tags that point at `images/positron-social.png`, the branded logo card, which returns HTTP 200.


juliasilge:
Closes https://github.com/posit-dev/positron-website/pull/444

jthomasmock:
Ah! I see see the positron-specific project _quarto.yml, now: https://github.com/posit-dev/positron-website/blob/e4c13cb12df63c41836ff6421afc9b98ca68261b/_quarto-positron.yml#L12

this LGTM!

jthomasmock (approved):

PR #446, opened 2026-08-07
Bump site to use Quarto 1.10.18

juliasilge:
This PR changes the pinned Quarto version from 1.9.23 to 1.10.18, in both `.quarto-version` and `netlify.toml`. The `positron` and `workbench` profiles both render without errors. 

What do we get with this bump?

- Quarto 1.10 gives us headings inside conditional content, which fixes our headings for 3 pages here. 
- It also improves our `llms-txt` output. Quarto 1.9 rewrote `.html` to `.llms.md` in every absolute URL, including links to other sites which corrupted 47 third-party links across 13 domains, for example `docs.posit.co/platform-support.llms.md`. In 1.10 that count is zero. This matters because these files ship as the `positron-llms` and `positron-workbench-llms` bundles, so any model that read them followed a 404.


midleman (approved):
Ran the docs-bundle workflow against this branch on staging (version `2026.08.0-331-quarto110`) and diffed the output against the 1.9.23 bundles from `2026.08.0-331`.

Nothing regressed: identical file lists in both profiles, `llms.txt` byte-identical, all 79 relative `.llms.md` links still resolve inside the bundle, same total word count, sizes unchanged.

It also fixes a bug we've been shipping. 1.9.23 was rewriting `.html` → `.llms.md` on absolute external URLs, so the current bundles contain 51 dead links (`docs.aws.amazon.com/.../getting-started-install.llms.md`, `ipython.readthedocs.io/.../magics.llms.md`, and so on). 1.10.18 leaves external URLs alone. 51 → 0.

Only other changes are cosmetic: hard line breaks moved from trailing-double-space to trailing backslash, and the `index.qmd` H1 renders as a setext heading now. Both still parse fine.

